PR #1, opened 2024-06-09
CI with preview

incognitojam:
- Regenerated lockfile (needed for pnpm v9, figured we didn't need to preserve it)
- Added CI workflows for deploying previews similar to connect
- Fixed type errors preventing build


incognitojam:
not sure why preview workflow didn't run! copied it from connect pretty much exactly. might be because it uses secrets

adeebshihadeh:
> not sure why preview workflow didn't run! copied it from connect pretty much exactly. might be because it uses secrets

Needs to be merged to master to run.

PR #4, opened 2024-06-09
fix: redirect to /login if can't fetch user profile

incognitojam:
This fixes the blank page you see if you haven't logged in


adeebshihadeh:
Preview workflow is working now, but I don't think we're pushing the built app correctly.

incognitojam:
thanks, will test it on #2

PR #7, opened 2024-06-09
switch to bun

incognitojam:


incognitojam:
```
Run cloudflare/pages-action@v1.5.0

Error: Input required and not supplied: apiToken
```

Dockerfile
```diff
@@ -1,47 +1,38 @@
-FROM node:18-alpine AS base
-
```
adeebshihadeh, line 2: we can remove all extra infra like this for now: Dockerfiles, nginx config, etc.

PR #12, opened 2024-06-10
added demo feature

pir8bay:


src/pages/auth/login.tsx
```diff
@@ -90,6 +96,22 @@ export default function Login() {
...
+        <Button
+          class="h-13 gap-4 hover:bg-gray-200"
+          onclick={loginAsDemoUser}
+          trailing={
+            <img
+              src="/images/chevron.svg"
+              alt=""
+              width={18}
+              height={18}
+              style={{ transform: 'rotate(180deg)' }}
```
ugtthis, line 109: Chevron facing right svg [link](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:chevron_right:FILL@0;wght@400;GRAD@0;opsz@20&icon.size=18&icon.color=%23e8eaed). Figured this would be better since the rest of the new-connect UI is also using Google Material Icons. Plus one less line of code since chevron faces correct way.

pir8bay: 👍

PR #20, opened 2024-06-12
CI(Preview): add retries to find PR script

ugtthis:
To my understanding this is so pose to help by increasing the rate limit since I added the Github Token. I also added a retry mechanism instead of trying to rerun actions manually. I noticed that Cameron also ran into the secondary rate limit so hopefully this would help us not hit that limit anymore.

I'm not entirely confident this would help so feel free to close this. My previous PRs are now able to generate the Preview Link.

Reference Link: [docs.github.com](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits)

### DIFF
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -30,18 +30,29 @@ jobs:
         uses: actions/github-script@v7
         with:
           script: |
-            const response = await github.rest.search.issuesAndPullRequests({
-              q: 'repo:${{ github.repository }} is:pr sha:${{ github.event.workflow_run.head_sha }}',
-              per_page: 1,
-            })
-            const items = response.data.items
-            if (items.length < 1) {
-              console.error('No PRs found')
-              return
+            async function findPR() {
+              for (let retries = 3; retries > 0; retries--) {
+                try {
+                  const response = await github.rest.search.issuesAndPullRequests({
+                    q: 'repo:${{ github.repository }} is:pr sha:${{ github.event.workflow_run.head_sha }}',
+                    per_page: 1,
+                  });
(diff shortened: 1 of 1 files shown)


ugtthis:
Meant to run this PR on my fork to see if it fixed the secondary rate limit problem I was hitting with my other PRs. My bad

.github/workflows/preview.yaml
```diff
@@ -30,18 +30,29 @@ jobs:
...
+                  console.error(`Retrying... (${3 - retries + 1}/3)`);
+                  if (retries === 1) throw error;
+                  await new Promise(res => setTimeout(res, 10000));
+                }
+              }
             }
-            const pullRequestNumber = items[0].number
-            console.info("Pull request number is", pullRequestNumber)
-            return pullRequestNumber
+            return findPR();
```
incognitojam, line 54: Is there a reason to wrap this in a function?

ugtthis: Maybe wrapping it in a function is not needed here, but the intention was to make it easier to manage, to debug, and easier to simulate different scenarios to make sure the retry logic works as expected. But maybe it’s overkill? Let me know if you prefer it a certain way

incognitojam: I think it's better to just set `retries` on the action, it will handle the `Retry-After` header for us
https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api#handle-rate-limit-errors-appropriately

ugtthis: Gotcha. Thanks for the fix!

PR #26, opened 2024-06-13
CI: use cloudflare/wrangler-action

incognitojam:
cloudflare/pages-action hasn't been updated and is using deprecated Node16


incognitojam:
need to test preview from fork

PR #38, opened 2024-06-23
Post screenshots of preview sites in PR comments

signed-long:


adeebshihadeh:
I made a new repo to push artifacts to: https://github.com/commaai/ci-artifacts. There's a deploy key for that repo in the `CI_ARTIFACTS_DEPLOY_KEY` GitHub secret.

I think the workflow file has to already use the secret in master, so if you make a minimal PR that does nothing but "echo", I'll merge that first so you can work on this.

signed-long:
> I made a new repo to push artifacts to: https://github.com/commaai/ci-artifacts. There's a deploy key for that repo in the `CI_ARTIFACTS_DEPLOY_KEY` GitHub secret.
> 
> I think the workflow file has to already use the secret in master, so if you make a minimal PR that does nothing but "echo", I'll merge that first so you can work on this.

Using the secret in master: https://github.com/commaai/new-connect/pull/40

incognitojam:
I think you should be able to use the secret now

signed-long:
So still can't run the workflow in a PR from a fork, or else anyone could then dump the secrets. But I ran it on my fork with some slight changes to point to my `ci-artifacts` repo etc. 

This is the comment it produces: https://github.com/signed-long/new-connect/pull/4#issuecomment-2193369487
This is the diff comparing this change to what created that comment: https://github.com/signed-long/new-connect/pull/6/files

@adeebshihadeh, before merging will also need the https://github.com/commaai/ci-artifacts repo initialized. This PR pushes the images to `new-comma/pr-<pr number>` on main. But can adjust to however you want to structure that repo.

PR #42, opened 2024-06-26
Dashboard UI overhaul

vishalkrishnads:
This is the implementation of a new dashboard UI for connect. It takes inspiration from messaging apps like Discord, Telegram, WhatsApp, etc.

> ⚠️ initially, you might have to refresh a couple of times and/or switch devices for the routes to load

https://github.com/commaai/new-connect/assets/50231856/d87917bd-abac-4c9d-ad90-1aedc161122d

## Features
* 70/30 split view in larger screens for routes/drive. The view itself is resizable, simply drag to resize.
* Search by place
* Basically everything from the old UI design

## Motivation
The current design seems to be wasting some space on larger screens.

## Status 
This is a proof of concept of this UI design. I'm open to feedback and suggestions.


sshane:
The 3D map is really cool! Is that easy to split out into a separate PR?

PR #54, opened 2024-07-07
Show device location on map

### LINKED ISSUE #47: Show device location on map
Must have these features from connect:
* comma device location shown
* user's phone/PC location shown
* locations are clickable for more info (reverse geocoding) and "open in maps"

Looks like this in current connect:
image

vishalkrishnads:
Fixes #47. I couldn't test the device location with demo account since the device doesn't report location at `/v1.1/devices/:dongleId`, it just returns `null`. User location works, and the logic is same for device location as well, so it should work as well.


adeebshihadeh:
CI didn't run for some reason - maybe due to the merge conflict. Can you rebase?

adeebshihadeh:
Can you rebase?

package.json
```diff
@@ -52,7 +52,9 @@
     "clsx": "^1.2.1",
     "dayjs": "^1.11.11",
     "hls.js": "^1.5.11",
-    "solid-js": "^1.8.17"
+    "mapbox-gl": "^3.5.1",
+    "solid-js": "^1.8.17",
+    "solid-map-gl": "^1.11.3"
```
adeebshihadeh, line 57: why do we need two new dependencies? what are they for? how big are they?

src/utils/device.ts
```diff
@@ -1,6 +1,7 @@
 import type { Device } from '~/types'
 
-export function getDeviceName(device: Device) {
+export function getDeviceName(device: Device | undefined) {
+  if(!device) return ''
```
adeebshihadeh, line 4: should be split into its own PR

package.json
vishalkrishnads: mapbox-gl is for the map itself. The static map on the cards is actually an image retrieved from the mapbox api and can't be interacted with like a normal map. solid-map-gl is a solid plugin that is essentially a wrapper around mapbox-gl that helps keep the code concise. mapbox-gl can be used standalone, but the diff would be too large. As for how large they are, mapbox-gl is a bit large (more here https://github.com/mapbox/mapbox-gl-js) and solid-map-gl is small (https://github.com/GIShub4/solid-map-gl). If you're okay with the larger diff, solid-map-gl can be removed, but mapbox-gl can't. However, if you're okay with the map not being interactive, I might be able to do it with zero additional deps.

src/utils/device.ts
vishalkrishnads: okay will do

adeebshihadeh:
This looks nice and tiny: https://leafletjs.com/

vishalkrishnads:
> This looks nice and tiny: https://leafletjs.com/

yes it does! switched to leaflet now. The bundle size check is still failing saying the size is `732K`, maybe the limit should be increased?

vishalkrishnads:
I'm unable to resolve the merge conflicts here. It says I don't have write access. Please look into it.

src/components/RouteVideoPlayer.tsx
```diff
@@ -1,8 +1,8 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call */
```
darioalessandro, line 1: why did you have to disable the linter?

PR #55, opened 2024-07-07
Dashboard: default to last selected device

### LINKED ISSUE #45: Implement same device sorting and selecting as connect
Current connect consistently sorts devices and remembers the last one you had selected

incognitojam:
Write the dongle ID to local storage when a device is selected, and use that as the default when loading the app (unless a dongle is already set in the URL)

closes #45


incognitojam:
Device sorting is already implemented here: https://github.com/commaai/new-connect/blob/cc11f95348d92f78b8d6527430b944b68758f78e/src/api/devices.ts#L5

PR #62, opened 2024-07-11
CI: bundle size limit

### LINKED ISSUE #57: Implement bundle size limit in CI
500K limit?

signed-long:
Checks if the gzip size from `vite-bundle-visualizer` is under 500K. Also uploads the html to the build artifacts. 

https://github.com/signed-long/new-connect/actions/runs/9900858794

Resolves #57


adeebshihadeh:
I like how simple this is, but might be worth using something like this https://www.npmjs.com/package/vite-bundle-visualizer

I ran that, and it made it super clear that HLS is most of the size. List of top 10 things with their size and percent of the total might be enough, though.

adeebshihadeh:
After reviewing the recent new PRs (particularly #68), I realized you were right with the first approach. Simple, readable bash script beats several-word-npm-dependency every time. Let's be very careful with our dependencies.

Want to make a PR to revert back to that and maybe make it a bit easier to see that HLS.js was the biggest thing?

signed-long:
> After reviewing the recent new PRs (particularly #68), I realized you were right with the first approach. Simple, readable bash script beats several-word-npm-dependency every time. Let's be very careful with our dependencies.
> 
> Want to make a PR to revert back to that and maybe make it a bit easier to see that HLS.js was the biggest thing?

#72 doesn't do anything wrt the seeing the deps size breakdown. But will follow up with that,  it'll just require parsing some json from the build step.

PR #65, opened 2024-07-12
Implement prime settings

incognitojam:
- prime
  - [x] show plan name and details
    - [x] show pending state (`getStripeSession`)
  - [x] update payment method
  - [x] cancel prime
    - [x] confirmation dialog instead of immediate cancel
- no prime
  - [x] show available plans
  - [x] go to stripe checkout
    - [x] show cancelled infobox

|checkout|stripe|
|---|---|
| prime checkout selected plan | prime_checkout_stripe |
| waiting for payment | processing |
| prime_manage_success_payment | prime_manage_success_processing |
| activated | cancel |
| prime_manage_activated | prime_manage_cancel |

| cancel large screen |
| -- |
| prime_manage_cancel_large |


adeebshihadeh:
Fixed CORS on the billing (and athena) server for the deployment previews!

ugtthis:
I noticed this one didn't have a design yet, wanted to send some ideas.

I noticed in current connect there are 2 settings button, seems like its because each device has its own little settings. Instead of having 2 separate settings btns, you can switch devices with a dropdown in the account settings page. This would then update the settings page with the relevant info for the chosen device from the dropdown.

The design below doesn't show all the pages but shows the general paths to the different things current connect already offers. I think new-connect should have all settings in one place so it's easier to navigate and find the information you need.

Current design has some space below the Subscriptions settings page which would prob be taken up with a comparison table of the different connect plans if this mock up was for a free plan.


**This general settings design has a path for:**
- a way to change devices via dropdown
- unpairing device
- uploads queue
- dongle id
- user id
- profile picture
- managing devices
- support (a page filled with helpful links, and FAQs)

prime settings 
- device name, current plan, when you joined, next payment, and amount
- Update payment button
- Cancel subscription
<br>

## Design v1
connect-settings-idea

PR #68, opened 2024-07-12
PWA Support

seanaguinaga:
https://new-connect-beta.vercel.app/

Add to home screen

turn on airplane mode to see offline fallback - not sure what you exactly want with this?

Assets are all generated from the two logo svgs - run build to see them in the dist folder


vercel.json
```diff
@@ -0,0 +1,3 @@
+{
```
adeebshihadeh, line 1: also unrelated?

package.json
```diff
@@ -21,6 +21,11 @@
   "type": "module",
   "devDependencies": {
     "@solidjs/testing-library": "^0.8.8",
+    "@vite-pwa/assets-generator": "^0.2.4",
+    "vite-plugin-pwa": "^0.20.0",
+    "sharp": "0.32.6",
+    "sharp-ico": "0.1.5",
+    "workbox-window": "^7.1.0",
```
adeebshihadeh, line 28: what do these 5 new dependencies do?

adeebshihadeh (changes requested):

vercel.json
seanaguinaga: Will remove - was just for my testing on vercel

adeebshihadeh:
This PR is +100 lines net, including a bunch of new dependencies and configs. I suspect doing all the PWA stuff ourselves would be less than or equal to this.

package.json
seanaguinaga: These are required for the minimal config vite-pwa plugin. 

They are used generate the assets like splash screens, icons, etc. at build time and are not included in the applications bundle. 

There's an alternative way to use "resolutions" in package.json for the sharp and sharp-ico libs but I couldn't get them to work in this repo.

seanaguinaga: It's a declarative plugin instead of an imperative config.

seanaguinaga:
> This PR is +100 lines net, including a bunch of new dependencies and configs. I suspect doing all the PWA stuff ourselves would be less than or equal to this.

Adding the splash screen meta tags is 60 lines alone in the html
Listing all the device sizes, names etc. etc.

Maintaining a list of sizes of assets for new devices, configuring service workers, etc. is a lot of grunt work that is not something I think any team should really be focused on maintaining or generating. 

Much more than 100 lines unfortunately

PR #70, opened 2024-07-12
Show which comma devices are active

### LINKED ISSUE #63: Show which comma devices are active in `cars_list`
in old connect users can see which comma devices are being used in the list
image

new-connect doesn't have this feature
image

griffinbaker12:
Fixes #63.

Uses a very similar approach to the original connect app by adding a ```fetched_at``` property after fetching the device data.

Sample UI Views:

image

image


src/types.d.ts
```diff
@@ -23,6 +23,10 @@ export interface Device {
   sim_id: string
 }
 
+export interface DeviceWithFetchedAt extends Device {
```
adeebshihadeh, line 26: why not just add to Device?

src/api/devices.ts
```diff
@@ -16,11 +16,22 @@ const sortDevices = (devices: Device[]): Device[] => {
   })
 }
 
+const populateFetchedAt = (device: Device): DeviceWithFetchedAt => {
+  return {
+    ...device,
+    fetched_at: Math.floor(Date.now() / 1000),
```
adeebshihadeh, line 22: this is kind of nice to know for everything. maybe make `fetched_at` generic?

src/utils/device.ts
```diff
@@ -1,6 +1,11 @@
...
 
-export function getDeviceName(device: Device) {
+export function getDeviceName(device: DeviceWithFetchedAt) {
   if (device.alias) return device.alias
   return `comma ${device.device_type}`
 }
+
+export function deviceIsOnline(device: DeviceWithFetchedAt) {
+  if (!device.last_athena_ping) return false
+  return device.last_athena_ping >= (device.fetched_at - 120)
```
adeebshihadeh, line 10: more readable to combine these imo

adeebshihadeh (changes requested):

src/types.d.ts
griffinbaker12: to keep the api's `DeviceType` distinct from the representation in the app

adeebshihadeh: i don't think that's too important, maybe just add an inline comment that that's not populated by the api?

griffinbaker12: cool, that works

src/api/devices.ts
```diff
@@ -16,11 +16,22 @@ const sortDevices = (devices: Device[]): Device[] => {
...
+const populateFetchedAt = <T>(item: T): WithFetchedAt<T> => {
+  return {
+    ...item,
+    fetched_at: Math.floor(Date.now() / 1000),
+  }
+}
+
 export const getDevice = async (dongleId: string) =>
   fetcher<Device>(`/v1.1/devices/${dongleId}/`)
+    .then(populateFetchedAt)
```
adeebshihadeh, line 28: can `fetcher` always do this?

src/types.d.ts
```diff
@@ -21,8 +21,12 @@ export interface Device {
   device_type: string
   openpilot_version: string
   sim_id: string
+  // not populated by api
+  fetched_at: number
```
adeebshihadeh, line 25: isn't this redundant with `WithFetchedAt`?

griffinbaker12: definitely now with fetcher returning this for all cases

src/api/devices.ts
griffinbaker12: yep, nice idea. assuming the api only returns objects and arrays of objects this should work. in the case its a primitive, I throw an error, but lmk and I can change.

griffinbaker12: I had fetcher return `<WithFetchedAt<T>>` instead of the union of `<WithFetchedAt<T>>` and `<WithFetchedAt<T>[]>`, so that we didn't have to perform additional checks throughout the project to determine whether we were dealing with an array or not.

PR #74, opened 2024-07-13
Use hls light

signed-long:
hls.js has a light build that removes some mostly audio related functionality.

From https://github.com/video-dev/hls.js#build-tasks:
>**NOTE:** `hls.light.*.js` dist files do not include alternate-audio, subtitles, CMCD, EME (DRM), or Variable Substitution support. In addition, the following types are not available in the light build:
>- `AudioStreamController`
>- `AudioTrackController`
>- `CuesInterface`
>- `EMEController`
>- `SubtitleStreamController`
>- `SubtitleTrackController`
>- `TimelineController`
>- `CmcdController`

Size difference:
```
RouteVideoPlayer-BW7fAlsk.js                     387.47KB
   hls.js/dist/hls.mjs                           
RouteVideoPlayer-CWnPZdkL.js                     289.48KB
   hls.js/dist/hls.light.min.js
```


adeebshihadeh:
-100k, nice!

Is this already the minimal HLS we need or are there still some features we're not using?

adeebshihadeh:
Another thought - any way we can just use the HTML5 `<video>` element? What would it take to switch to that?

signed-long:
> -100k, nice!
> 
> Is this already the minimal HLS we need or are there still some features we're not using?

Yeah this is using the minimal version 

> Another thought - any way we can just use the HTML5 `<video>` element? What would it take to switch to that?

It's a browser support thing, some browsers can take the `.m3u8` files and support hls streaming in the `<video>` tag natively. The hls library lets unsupported browsers ([most](https://caniuse.com/?search=hls)) able to stream them too.

adeebshihadeh:
I meant, can we strip down their light build even more for our use case?

PR #76, opened 2024-07-13
Fixes engage %

### LINKED ISSUE #75: `engage %` is not correct
It was quick to realize when looking at the timeline that the engagement percentage did not match up. Grey represents overrides by driver such as pressing on the gas or moving the steering wheel.

My understanding is it's marking one state as two (engaged AND overriding) here's the place where the percentage is being calculated.

https://github.com/commaai/new-connect/blob/84d1b82e08e0b985a0e2ffa8ec879d2065bb5e47/src/components/RouteStatistics.tsx#L9-L13

This is where the state is determined (engaged, overriding, alert, userflag)

https://github.com/commaai/new-connect/blob/84d1b82e08e0b985a0e2ffa8ec879d2065bb5e47/src/api/derived.ts#L126-L171

image

griffinbaker12:
Fixes #75.

Ensures that the ```engaged``` and ```overriding``` states are mutually exclusive. The old logic would attribute both these states to ```engaged```, leading to inflated stats.


adeebshihadeh:
This should probably just come from the API, so all the logic is in one place. I'll have to look into that and come back to this.

BBBmau:
Looks to be better but like @adeebshihadeh said it would be best to get this directly from the API rather than calculating it on the client side.

griffinbaker12:
> 
> 
> Looks to be better but like @adeebshihadeh said it would be best to get this directly from the API rather than calculating it on the client side. 

yeah agreed, was cool to solve nonetheless

incognitojam:
We're looking to add this data to the API instead

PR #83, opened 2024-07-14
CI: Call preview from build workflow

### LINKED ISSUE #80: CI: show preview job status in PRs
The `preview` job which deploys PRs and posts the screenshots runs as a `workflow_run` GitHub Actions job, so its status only shows up in the Actions tab. This means it can silently break.

I gave it a quick attempt in https://github.com/commaai/new-connect/pull/79, but that action is broken and not maintained anymore.

signed-long:
It's possible to write the status back to the PR using the github checks api, but I think this is simpler.

Resolves #80

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -52,6 +52,7 @@ jobs:
           command: pages deploy dist --project-name=connect --branch=new-connect --commit-dirty=true
   
   preview:
+    needs: build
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Got an example PR for validation?

adeebshihadeh:
Ah found it on your fork. Make sure to include it in the description next time.

Issue #59, opened 2024-07-11
Implement PWA features


adeebshihadeh:
This issue is for making this a complete PWA, including but not limited to:
* icons
* splash screen
* a defined offline mode (doesn't have to do anything, except say "connect offline")

As well as testing the PWA-ness in CI.

adeebshihadeh:
This looks interesting https://github.com/elegantapp/pwa-asset-generator

seanaguinaga:
How's this: https://github.com/commaai/new-connect/pull/68

griffinbaker12:
design wise for the exit screen I had worked on something like this:

![Image](https://github.com/user-attachments/assets/d87f81d3-286e-4763-806b-f4c28be254e2)

theVedanta:
Hello,
Is this issue closed? I would like to try this out

namnbk:
Is this still available?

griffinbaker12:
anything not closed is open, but they aren't focusing on this rn

pedrosimao:
Hey, I see there are two PRs open since October 2024. Is it still open for a bounty? And why did the PRs did not get merge?

Eunanibus:
I'm also curious. If the work to combat these bounties isn't being merged, I don't want to tackle in-progress tasks without understanding why.

PR #92, opened 2024-08-01
pwa support

griffinbaker12:
addresses #59.

adds pwa support to the new connect app with no external dependencies. 

here is a [video](https://gyazo.com/b46ce751dea3928defc5f8d2da63c2a1) of it in action and a screenshot of the `offline.html` when connection is lost:

image


griffinbaker12:
seems like the new assets pushed the bundle over the limit

PR #93, opened 2024-08-03
chore: sidebar defaults to open on large screens

### LINKED ISSUE #73: Always show sidebar on PC

35C4n0r:
- If the screen Width is greater than 1024px, the sidebar defaults to open
- Removed the opacity overlay on sidebar open.

Closes #73


adeebshihadeh:
Why is the back button still there? It should be hidden.
image

PR #99, opened 2024-08-19
fix: sort routes without start_time/end_time

### LINKED ISSUE #60: Handle routes with no time
Since openpilot 0.9.7 [removed RTC support](https://blog.comma.ai/097release/#new-route-format), some routes can have no time (e.g. the device just booted up and didn't get a GPS fix during the route).

Since openpilot's uploader guarantees that routes are uploaded in the order they were created, we can still show all routes in the correct sequence using the `create_time` attribute of the route.

Requirements
* hide the date/time
* show "timeless" routes in the correct order with the rest of routes

tex-murphy:
- falls back to `create_time` when start_time/end_time are unavailable
- refactor `<RouteHeader />` into its own file
- hide `<RouteStatistics />` if there's no trip data

closes #60

### DIFF
--- a/src/components/RouteCard.tsx
+++ b/src/components/RouteCard.tsx
@@ -1,33 +1,12 @@
 import { Suspense, type VoidComponent } from 'solid-js'
-import dayjs from 'dayjs'
 
-import Avatar from '~/components/material/Avatar'
-import Card, { CardContent, CardHeader } from '~/components/material/Card'
-import Icon from '~/components/material/Icon'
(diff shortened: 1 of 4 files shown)


tex-murphy:
screenshot

Screenshot 2024-08-18 at 7 48 22 PM

tex-murphy:
@incognitojam @adeebshihadeh I believe this should squash web bounty #60 

any suggestions for moving this forward -- I can add mock data for this case if there's a good place to put that.

PR #108, opened 2024-10-20
Pwa deployment in cloudflare

knownotunknown:
## Features
- Caches core assets for offline use
- Shows offline page when network fails

## Testing
- Added Lighthouse PWA testing in CI
- Requires perfect PWA score (1.0) to pass CI checks
- Tests core PWA requirements including:
  - Service worker registration
  - Valid manifest
  - Required icons
  - Proper display settings
  
## Potential Improvements
- Can implement the [Richer PWA UI](https://web.dev/patterns/web-apps/richer-install-ui) which allows us to customize the PWA installation process
  
  ## Demo (Ctrl + r to reload PWA app)
  
https://www.loom.com/share/042bfb4cdda84da0b4775f44cfd4ff05?sid=7f6a27a2-5de4-4d17-bb9f-d7984d4d0bd2

Closes [#59](https://github.com/commaai/new-connect/issues/59)

PR #110, opened 2024-11-04
feat: added PWA asset generation

knownotunknown:
## Features
- Added and ran package.json script to handle PWA asset generation and integration
- Used https://github.com/elegantapp/pwa-asset-generator for PWA asset generation

## CI
- Excluded PWA assets from bundle size calculation

PR 1 of 2 to close https://github.com/commaai/new-connect/issues/59. 

2nd PR will include caching via service worker and testing via Lighthouse GitHub Actions. Could split up caching and testing if needed, but not sure if that's worth it/makes sense.

Original, full PR: https://github.com/commaai/new-connect/pull/108


index.html
knownotunknown, line 1: All changes in this file are auto-generated

public/manifest.json
knownotunknown, line 1: All changes in this file are auto-generated

package.json
```diff
@@ -11,6 +11,9 @@
   "scripts": {
     "build": "vite build",
     "dev": "vite",
+    "generate-pwa-assets": "npx pwa-asset-generator ./public/images/icon-vector.svg ./pwa-assets -m ./public/manifest.json -i ./index.html --sizes \"72x72,96x96,128x128,144x144,152x152,192x192,384x384,512x512\" --quality 75 --background \"#ffffff\" --padding \"10%\" --platform android,ios --opaque false",
+    "move-pwa-assets": "rm -rf ./public/pwa-assets && mv ./pwa-assets ./public/",
+    "pwa-setup": "npm run generate-pwa-assets && npm run move-pwa-assets",
```
knownotunknown, line 16: 1. "generate-pwa-assets":
   - Takes our vector icon from public/images/icon-vector.svg
   - Generates various sizes of icons required for PWA (72px to 512px)
   - Updates manifest.json and index.html with the new icons
   - Applies settings:
     * 75% image quality for optimization
     * White background
     * 10% padding around icons
     * Compatible with both Android and iOS
     * Transparent background allowed

2. "move-pwa-assets":
   - Moves generated assets to the public directory
   - Cleans up old assets first to avoid duplicates

3. "pwa-setup":
   - Runs both commands in sequence to generate and organize PWA assets

index.html
knownotunknown, line 1: All changes in this file are auto-generated

public/manifest.json
knownotunknown, line 1: All changes in this file are auto-generated

package.json
```diff
@@ -11,6 +11,9 @@
   "scripts": {
     "build": "vite build",
     "dev": "vite",
+    "generate-pwa-assets": "npx pwa-asset-generator ./public/images/icon-vector.svg ./pwa-assets -m ./public/manifest.json -i ./index.html --sizes \"72x72,96x96,128x128,144x144,152x152,192x192,384x384,512x512\" --quality 75 --padding \"10%\" --platform android,ios --opaque false",
```
adeebshihadeh, line 14: let's document these in the README?

public/manifest.json
adeebshihadeh: is the whole file auto generated?

adeebshihadeh (changes requested):
Can we gitignore the generated assets and just generate them at build-time?

public/manifest.json
knownotunknown: No, only the changes shown.

knownotunknown:
> Can we gitignore the generated assets and just generate them at build-time?

Sure. I went ahead and created separate build and build:prod commands since generating pwa-assets increases total build time from 1s to 5s, and you only need to generate pwa-assets once anyways.

adeebshihadeh:
> Sure. I went ahead and created separate build and build:prod commands since generating pwa-assets increases total build time from 1s to 5s, and you only need to generate pwa-assets once anyways.

This is not a good tradeoff. What happens if the assets get updated? How do I discover this `build:prod` command? It should just be part of the normal build. One case is 10x better than two cases.

knownotunknown:
> > Sure. I went ahead and created separate build and build:prod commands since generating pwa-assets increases total build time from 1s to 5s, and you only need to generate pwa-assets once anyways.
> 
> This is not a good tradeoff. What happens if the assets get updated? How do I discover this `build:prod` command? It should just be part of the normal build. One case is 10x better than two cases.

That's fair. I wasn't sure about the reasoning for .gitignoring the pwa-assets since those assets would probably rarely change anyways, and it would prevent an extra build step, so I thought we were really just optimizing for cleanliness. 

Was the main motivation of .gitignoring just to keep it our assets up-to-date with the source svg?

adeebshihadeh:
Yes, this way they're never out of date and the SVG is the source of truth.

PR #113, opened 2024-11-29
bug: CSS fix for larger screens

0x41head:
Changes:
- fixes the ugly CSS for larger screens


incognitojam:
It's not clear from the diff what CSS changed, feel free to reopen with more specific changes

PR #114, opened 2024-12-13
Show device location on map

### LINKED ISSUE #47: Show device location on map

leoforney:
Fixes #47. Was able to mock and test device location data. Note, there is a bug in leaflet that requires a setTimeout and invalidateCache call after initialization. Without this, only a tile in the top left corner will load.


leoforney:
https://github.com/user-attachments/assets/e07fe649-6567-43e7-9ce7-6eb9bd6dc709


https://github.com/user-attachments/assets/d9cad9e9-9c03-4549-9cd2-35424c960ca1

Demo videos

PR #115, opened 2024-12-13
feat: show user and device location on map

jtgi:
fixes: https://github.com/commaai/new-connect/issues/47

notes:
- fixes & evolves https://github.com/commaai/new-connect/pull/54 from @vishalkrishnads (🥂)
- requesting location perms: current connect requires clicking map (invisible), this pr adds a button. shows first time only.
- testing: don't have a device yet but mocked the api in the vid below.

https://github.com/user-attachments/assets/e514a7fe-fcc3-4d90-9be8-e38cc9ff3670


incognitojam:
This looks nice! I don't think the `last_gps_lat` and `last__gps_lng` are ever populated in the API though, old connect uses [`GET /v1/devices/:dongle_id/location`](https://api.comma.ai/#device-location) instead

jtgi:
> This looks nice! I don't think the `last_gps_lat` and `last_gps_lng` are ever populated in that part of the API though, old connect uses [`GET /v1/devices/:dongle_id/location`](https://api.comma.ai/#device-location) instead

Nice catch, ty. Updated to use that endpoint.

Quick note though, the endpoint 403s. I assume its because of the demo account since old connect does as well.
Works fine when I mock though and based on [comma-api  spec[(https://github.com/commaai/comma-api/blob/master/openapi.yaml#L183-L215) I expect it to work.

Got a device you can try it with? Def need to get one setup or mocked.

incognitojam:
Yes it works for me :slightly_smiling_face:
image

I think not having a location for the demo car is OK for now - there's an argument that should be added on the backend the same way the demo routes are available. Would be nice to have the option to mock different data in the future too.

I noticed the "Show my location" button doesn't work on Firefox. It asks me to allow access to my location, but after that no marker for my location appears on the map. There's no error in the console either. It works fine in Chrome.

jtgi:
> Yes it works for me 
nice, thanks.

> I noticed the "Show my location" button doesn't work on Firefox. It asks me to allow access to my location, but after that no marker for my location appears on the map. There's no error in the console either. It works fine in Chrome.
Thanks, fixed this and learned a bit about browser implementation differences with geolocation.

Key difference: Safari and Firefox only acknowledge the permission as granted and fires events if you explicitly check 'Remember this for next time'. If the user allows for the current session only, change events don't fire and the permission will still return as 'not granted'. Chrome handles as I expected though.

Fix was to consume the position directly from the `navigation.geolocation.getCurrentPosition` call and not strictly rely on change event.

jtgi:
@incognitojam lemme know if you have any other feedback, happy to fix up

incognitojam (changes requested):

src/map/index.ts
```diff
@@ -6,13 +6,14 @@ import {
...
 } from './config'
+import { getThemeId } from '~/theme'
 
 export type Coords = [number, number][]
 
 const POLYLINE_SAMPLE_SIZE = 50
 const POLYLINE_PRECISION = 4
 
-function getMapStyleId(themeId: string): string {
+export function getMapStyleId(themeId: string): string {
```
incognitojam, line 16: I don't think this needs to be exported anymore

jtgi: yep, removed

incognitojam (approved):

incognitojam:
Thanks for the PR! You can open a ticket at `https://comma.ai/support?topic=Shop%20Order` to claim your bounty, just provide a link to this PR

jtgi:
@incognitojam awesome – thanks for the reviews and touch ups.

PR #117, opened 2024-12-31
feat(RouteList): implement infinite scrolling

LouisVelez:
This PR improves the RouteList component by introducing infinite scrolling and a responsive grid layout. These changes enhance user experience, performance, and maintainability while reducing manual interactions.

### DIFF
--- a/src/pages/dashboard/components/RouteList.tsx
+++ b/src/pages/dashboard/components/RouteList.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-misused-promises */
 import {
   createEffect,
   createResource,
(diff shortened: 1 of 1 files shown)


LouisVelez:
@adeebshihadeh I'm leaving this in draft since it was whipped up quickly as a POC. If this is of interest I can refactor the code, add test and submit the PR.

incognitojam:
I like the look of this and it looks pretty simple too

sshane:
It only loads 6 routes on desktop, so it doesn't fill up the space, and you can't scroll to load more. Other than that, this is nice!

incognitojam:
I reverted the style changes (grid layout) because we might be removing the map from the RouteCard and a list would be better for showing all of the routes

PR #124, opened 2025-01-05
Show Device & User location on map

derppey:
This fixes: https://github.com/commaai/new-connect/issues/47

**Notes:**
- User & device location are clickable and displayed on the map
- Mocked device location for testing

**Screenshots:**
screenshot-1
screenshot-2


incognitojam:
Thanks for submitting a PR, but in the end I went with another option

Issue #134, opened 2025-01-15
Log out buttons squished


sshane:
![image](https://github.com/user-attachments/assets/30972ac6-a205-45b6-9c80-6a1b54134de3)

Issue #82, opened 2024-07-14
Visual bugs with lots of devices


adeebshihadeh:
Sign out gets squished and the background color changes.

![image](https://github.com/user-attachments/assets/be4395ee-658e-4c3d-a267-3170928bc993)

PR #135, opened 2025-01-17
fix: buttons squished

mikhailmtion:
This resolves issue https://github.com/commaai/new-connect/issues/134 and #82 

**Before**
before

**After**
after


incognitojam:
When the list of devices doesn't fit and you scroll in the drawer, the background height seems to be limited to the screen height (I reproduced this by duplicating the `<For each={props.devices}>` in `DeviceList`)
image

Think you can fix that?

mikhailmtion:
@incognitojam Added scroll
Maybe we'll add a margin between the ListItem?
scroll

Issue #133, opened 2025-01-15
Add max-width for route viewer


sshane:
Too large on desktop screens

![image](https://github.com/user-attachments/assets/bf28b035-bf8c-4e45-879b-21052c982147)

With

```
    max-width: 800px;
    margin-left: auto;
    margin-right: auto;
```

![image](https://github.com/user-attachments/assets/9f901cee-74fe-40d8-adf6-734b2aca45fe)

PR #136, opened 2025-01-17
fix(dashboard): set content max-width

mikhailmtion:
This resolves issue https://github.com/commaai/new-connect/issues/133

**1280px**

**3840px**


incognitojam:
Can you try putting this in [`Dashboard.tsx`](https://github.com/commaai/new-connect/blob/master/src/pages/dashboard/Dashboard.tsx#L103) (wrap the `<Switch>` in a div) so that it applies to every page?

mikhailmtion:
@incognitojam Yes, sure

image

PR #103, opened 2024-09-23
Feat(route): add ability to control video via the timeline

t1mp4:
As suggested by https://github.com/commaai/new-connect/issues/64. Handles both mouse and touch input. Tested on iPhone 14.

PR #139, opened 2025-01-21
Feat: draggable video timeline

mikhailmtion:
Implemented functionality for dragging and dropping videos using a timeline
The old version of **connect.comma** a has this
With support mobile

**Result:**

dndTimeline-ezgif com-video-to-gif-converter (1)


incognitojam:
Thanks for the PR, but this was already implemented in #103

PR #141, opened 2025-01-22
fix(PairActivity): video container layout

mikhailmtion:
**Before**
before

**After**
after


incognitojam:
On smaller screens the page can scroll horizontally

PR #142, opened 2025-01-22
(patch) Disable vertical scrolling when dragging across the timeline

t1mp4:
Patch for #103


incognitojam:
Fixed with CSS instead https://github.com/commaai/new-connect/pull/143

PR #144, opened 2025-01-30
UI bug: Unintended movement when zooming in or out

ugtthis:
<table>
  <tr>
    <th>Current UI</th>
    <th>Proposed UI</th>
  </tr>
  <tr>
    <td>Current UI</td>
    <td>Proposed UI</td>
  </tr>
</table>

- The vertical scrollbar is shifting because the content width calculation doesn't account for the scrollbar width when using VW(w-screen)
- UI movement happens when the drawer is open and closed, and when viewing a route
- Safari is the browser that only shows movement when the drawer is open

<br>
<br>

# Testing bug on different browsers

## Firefox
firefox

<br>

## Google Chrome
chrome

<br>

## Safari - only shows the movement when drawer is open
safari

### DIFF
--- a/src/components/material/Drawer.tsx
+++ b/src/components/material/Drawer.tsx
@@ -35,7 +35,7 @@ const Drawer: ParentComponent<DrawerProps> = (props) => {
       </nav>
 
       <main
-        class="absolute inset-y-0 w-screen overflow-y-auto bg-background transition-drawer duration-500"
+        class="absolute inset-y-0 w-full overflow-y-auto bg-background transition-drawer duration-500"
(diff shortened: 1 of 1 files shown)


ugtthis:
Noticed that with this current change, Safari is still having the same resize motions when sidebar is open. I think it has to do with the `transition-drawer` and how Safari treats it differently than the other browsers.

I'm currently working on updating the always-on-sidebar PR in which this unintended motion would not exist. So currently this PR does fix Chrome and Firefox but Safari experience stays the same.

ugtthis:
Closing this PR because of #161

PR #146, opened 2025-01-31
UI:  Always show drawer when on desktop(updated)

ugtthis:
- text logo is clickable and takes you back home
- `<main>` content is responsive and drawer will pull out if screen is wide enough
- added warning icon when there is "no device" in drawer
- resolves issue #73 

## Proposed UI 📹
https://github.com/user-attachments/assets/56e706a2-4315-4e60-b520-f3082f9f31a8

### DIFF
--- a/src/components/TopHeader.tsx
+++ b/src/components/TopHeader.tsx
@@ -15,7 +15,15 @@ const TopHeader: Component = () => {
             menu
           </IconButton>
         )}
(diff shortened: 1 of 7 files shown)


incognitojam:
This does resolve the issue but I don't think the style changes are an improvement... I changed Drawer to handle its own state (https://github.com/commaai/new-connect/pull/156) and it made this not too difficult (https://github.com/commaai/new-connect/pull/157)

PR #100, opened 2024-08-26
UI: Route card expands to show quick actions

ugtthis:
Decided on this design versus my [popover design](https://github.com/commaai/new-connect/issues/91#issuecomment-2300172033) because this was a more straight forward, and had less complex UI layout/logic. Only slightly unrelated change I did was create a `RouteMap` component which made the return JSX  of `RouteCard.tsx` easier to read. Photos below are up to date with recent commits

This resolves issue #91. 

## Added
- Preserve/unpreserve Routes
- Enable/disable Public Access
- Copy Route ID
- View RouteID
- Ability to view route in useradmin
- Error when toggle doesn't update

<br>

## Card expanded
PR-100-screenshot

<br>

## Interacting w/UI changes while showing network tab:
https://github.com/user-attachments/assets/ecd27cf2-89e1-4fa6-990e-171fef263243

<br>

## Error state of toggle
errorState-PR--100

PR #164, opened 2025-02-04
RouteInfo: add route action buttons

ugtthis:
Moved UI from PR #100 to [RouteActivity](https://github.com/commaai/new-connect/pull/100#issuecomment-2632021572)

- On demo account toggles will show error message. Toggles will work when you sign into an account

# Collapsed RouteInfo UI
collapsed-ui

# Demo of Proposed UI (updated)
RouteInfo-UI


incognitojam:
I'm still not sure this is the right place, especially as they take up a big chunk of screen space

Maybe the copy route ID and view in useradmin shortcuts should be there, but the preserve route/public access toggles should be hidden in a menu. For now, it might be okay to move those below the map.

What do you think?

adeebshihadeh:
I kind of like it actually. I find the current connect menus pretty annoying and often click the wrong one when trying to find something.

It's probably like 30% too big, but I like the general concept.

ugtthis:
I generally think the map isn’t the main thing users use. And the current connect has the map tucked away(which I prefer). I also haven’t ran into users on discord commenting about how much they use the map or how they want to see more of the map when clicking into the route.

The usual comments has to do with video playback, uploading, timeline, and general route info(preserved, route id, public, etc) So I favor the size and spacing of `RouteActions` currently since it will also feels/looks nice when being used on a iPhone SE(smaller width screens).

---

Though I do think we can save a lot on spacing if we structure the page a little different. Refer to video below.

Main differences:
- `Timeline` joins in w/video player
- Adjusted video progress bar to show closer in sync w/`Timeline`
- RouteStats becomes `RouteInfo` which also has a expand button to show the `RouteActions`

With these changes we get the size reduction by default without sacrificing the current spacing of text and buttons so that smaller width devices have a good UX also when using new-connect


# Rough demo
idea-1



## Comparing UI:
| Current UI | Proposed UI |
|------------|------------|
|  |  |

ugtthis:
Link to the above #190  RouteActivity: Design Idea demo :
### [https://190.connect-d5y.pages.dev/](https://190.connect-d5y.pages.dev/)

incognitojam (changes requested):

src/components/RouteActions.tsx
```diff
@@ -0,0 +1,153 @@
...
+        />
+        <ToggleButton
+          label="Public Access"
+          active={() => isPublicLocal()}
+          onToggle={() => void toggleRoute('public')}
+        />
+      </div>
+
+      <div class="mt-4 flex gap-[.75rem]">
+        <ActionButton
```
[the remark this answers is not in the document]
ugtthis: Made route id on 2 separate lines for smaller width phones. On one line the Route ID looks crammed or the font is too small. Thought the below change was the ideal version

## UI on small width screen
Screenshot 2025-02-13 at 7 05 16 PM

src/pages/dashboard/activities/RouteActivity.tsx
```diff
@@ -58,8 +81,19 @@ const RouteActivity: VoidComponent<RouteActivityProps> = (props) => {
...
             updateTime={onTimelineChange}
           />
-          <Suspense fallback={<div class="h-10" />}>
-            <RouteStatistics route={route()} />
+        </div>
+
+        <div class="flex flex-col gap-2">
+          <h3 class="text-label-sm uppercase">Route Info</h3>
+          <Suspense fallback={<div class="skeleton-loader min-h-32 rounded-lg bg-surface-container-low" />}>
+            <RouteInfo
```
incognitojam, line 89: I don't think passing "initial values" like this is effective at preventing Suspense as you are using the resource to get the initial values...

I would try moving the `isPublic` and `isPreserved` resources into the `RouteActions` component since nothing else uses those values, and that way the parent `Suspense` should be triggered

src/components/RouteInfo.tsx
```diff
@@ -0,0 +1,55 @@
...
+            isPreserved={props.isPreserved}
+          />
+        </Suspense>
+      </Show>
+
+      <button
+        class={clsx(
+          'flex w-full cursor-pointer justify-center p-2 hover:bg-black/45',
+          expanded() 
+            ? 'rounded-b-md border-2 border-t-0 border-surface-container-high bg-surface-container-lowest'
```
incognitojam, line 42: These classes can be simplified (`rounded-b-md bg-surface-container-lowest` in both cases)

incognitojam (changes requested):
Looks good so far, just see if we can cleanup passing signals around

src/components/material/Button.tsx
```diff
@@ -11,6 +11,7 @@ type ButtonProps = ButtonBaseProps & {
   loading?: boolean
   leading?: JSXElement
   trailing?: JSXElement
+  noPadding?: boolean
```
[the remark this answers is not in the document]
ugtthis: This was used for the old action btns, forgot to remove. Thanks for catching that!

src/pages/dashboard/activities/RouteActivity.tsx
ugtthis: That makes sense. Thanks for these suggestions!

src/components/RouteActions.tsx
```diff
@@ -0,0 +1,144 @@
...
+    <div class="flex flex-col border-2 border-t-0 border-surface-container-high bg-surface-container-lowest p-5">
+      <div class="font-mono text-body-md text-zinc-500">
+        <h3 class="mb-2 ml-2">Route ID:</h3>
+        <button
+          onClick={() => void copyCurrentRouteId()}
+          class="flex w-full cursor-pointer items-center justify-between rounded-lg border-2 border-surface-container-high bg-surface-container-lowest p-4 hover:bg-surface-container-low"
+        >
+          <div class="lg:text-body-lg">
+            {currentRouteId().split('/')[0] || ''}/
+            <br />
```
incognitojam, line 114: Is this break necessary?

ugtthis: Without it we will be decreasing the font even more when its a smaller screen or will run into responsive layout issues with the longer string. 

I tried wrapping the words to the new line when it gets to small but it would jump to the new line in unideal places as the screen gets smaller. I thought the natural break that also makes it easier to read the string would be at the `/` character.

### iPhone SE width screen
iphone-se-width

PR #190, opened 2025-02-12
RouteActivity: Design Idea

ugtthis:
Code needs work but this is the general idea. Refer to PR comment [here](https://github.com/commaai/new-connect/pull/164#issuecomment-2654845246).

- This would get split up into 2 PRs
- [First PR](https://github.com/commaai/new-connect/pull/164): `RouteInfo` with expand button that shows [RouteActions](https://github.com/commaai/new-connect/pull/164#issuecomment-2654845246)
- Second PR: `Timeline` and `RouteVideoPlayer` joining with the UI tweaks

PR #191, opened 2025-02-14
Fix: update geocoding test to correct address

ugtthis:
## Error
Screenshot 2025-02-13 at 9 15 28 PM

- Was getting the above error in PR #164 . Seems like the current address are now wrong because mapbox got more accurate, which then changed to the actual address of the white house. Below is the address of the white house also found [here](https://www.usa.gov/agencies/white-house).
- Got the same error when I did a blank commit in PR #190 
<br>

## White House address
Screenshot 2025-02-13 at 9 33 49 PM

### DIFF
--- a/src/map/geocode.spec.ts
+++ b/src/map/geocode.spec.ts
@@ -14,7 +14,7 @@ describe('getFullAddress', () => {
   })
 
   test('normal usage', async () => {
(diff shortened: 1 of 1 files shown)


incognitojam:
probably my fault for not checking the white house address lol

Issue #91, opened 2024-07-29
add ability to mark route as `preserved` / `public access`


BBBmau:
currently their is no way to preserve or mark a route for public access on new-connect, forcing me to move back to the old connect in order to do this.

from old-connect: 
![image](https://github.com/user-attachments/assets/64653f40-1aab-4eaa-ac91-dd6d61918a73)

Unsure how this could be added in design wise.

ugtthis:
Came up with a design that uses a small popup that is triggered by clicking an icon that is placed at the bottom right of the trip card. I wanted to keep the popup with the most used(in my opinion) options which is why I left out "View in useradmin"

I was debating between a popup or a bottom drawer, but was thinking that this popup design would possibly be less code and easier to reuse when it came to desktop also.

## Design
![routeItemPopup](https://github.com/user-attachments/assets/4ce0426c-4159-4399-aa46-ba51cca0f099)

PR #197, opened 2025-02-22
UI Rework: Route activity buttons

Cynamide:
Did a redesign of the UI components of route actions to match the overall aesthetic of the app. Issue: #91 

Unexpanded Before      |  Unexpanded After
:-------------------------:|:-------------------------:
unexpanded before   |  unexpanded after

Expanded Before      |  Expanded After
:-------------------------:|:-------------------------:
expand before | expand after

Error Before     | Error After
:-------------------------:|:-------------------------:
error before | error

### DIFF
--- a/src/components/RouteActions.tsx
+++ b/src/components/RouteActions.tsx
@@ -14,7 +14,7 @@ const ToggleButton: VoidComponent<{
   onToggle: () => void
 }> = (props) => (
   <button
-    class="flex w-full items-center justify-between px-5 py-2 transition-colors hover:bg-surface-container-low"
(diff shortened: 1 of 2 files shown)


incognitojam:
I like the direction of this but I think now the toggles and text are a little too small

Cynamide:
I see, I kinda had the same thinking too but I wanted to match it with the sizes we have already on screen. I could try something different but I noticed the issue is closed now. Does this mean that there wont be any changes to it and the PR should be closed too?

incognitojam:
No I closed that issue because the feature is implemented, I'm still open to iterating on the design

incognitojam:
Thanks but closing in favour of #212

Issue #181, opened 2025-02-08
Cache playwright install in GHA


incognitojam:
Should be able to cut CI time by 10 seconds

PR #199, opened 2025-02-23
Cached playwright in test job.

Cynamide:
Job now uses cached playwright files to run test. Key is generated from the hash of `bun.lockb` file.  Issue: #181 

I am not sure if this approach is better than directly having key as the playwright version contained in `package.json`. Any changes in the lock file will trigger `bun playwright install` even if there are no playwright specific changes in that file.

build / test reduced from 37secs ----> 19secs


incognitojam:
looks good to me, thanks!

PR #200, opened 2025-02-28
fix: view shared devices

incognitojam:
use test cases to validate redirect to login page and to view shared devices/public routes!

- you can view other devices even when you don't own them or are not signed in
- you can view public routes

:warning: video does not load for shared routes, yet


sshane:
@incognitojam this broke https://github.com/commaai/connect/issues/291

incognitojam:
ah sorry, I think it affects read-only users though

PR #203, opened 2025-03-15
feat: upload queue

greatgitsby:
second part of https://github.com/commaai/connect/issues/51


src/api/workflow.ts
```diff
@@ -0,0 +1,48 @@
...
+  fcamera: ["fcamera.hevc"],
+  dcamera: ["dcamera.hevc"],
+  ecamera: ["ecamera.hevc"],
+};
+
+export const uploadAllSegments = (routeName: string, totalSegments: number) => uploadSegments(routeName, 0, totalSegments - 1)
+
+export const uploadSegments = async (routeName: string, segmentStart: number, segmentEnd: number) => {
+    const dongleId = routeName.split('|')[0]
+    const routeId = routeName.split('|')[1]
```
greatgitsby, line 15: i don't like this behavior being duplicated throughout the code base. we can have a follow up issue to clean this up. but keeping it consistent in the meantime

greatgitsby:
how flexible are the LoC and bundle size numbers? i will do some refactoring but there will be limits.

incognitojam:
The hard limit is 500 kb, 5000 lines

I see about 200 lines is in the API fetchers, another 100 for types - for now I think you can save a bunch of lines by wrapping the lines less. I'd like to move this out to a separate package later if it's all standard comma API stuff (there is already a comma API package but it doesn't have types)

greatgitsby:
thanks!

yea moving api calls and types out will help quite a lot. i'll do some refactors and it should be achievable

incognitojam:
Need any help finishing this? You could also split the requesting uploads and upload queue into separate PRs

greatgitsby:
nope, i've been working on the clip bounty. i'm planning on separating the API contract and calls separately as that is well defined now

incognitojam:
Can you split this up at all? It's pretty difficult to review

greatgitsby:
let me see what i can do. some refactors were for DRY (like the stats bar, that code was already somewhere else)

greatgitsby:
removed the stats bar component and changes to the progress bar

incognitojam:
> let me see what i can do. some refactors were for DRY (like the stats bar, that code was already somewhere else)

> removed the stats bar component and changes to the progress bar

I just meant you could open PRs to fix/cleanup things first, which would be small and easy to merge, if they help make the diff for this PR smaller

greatgitsby:
created these which are self-contained:

https://github.com/commaai/connect/pull/235
https://github.com/commaai/connect/pull/236
https://github.com/commaai/connect/pull/237

i'll do some component refactors later today

package.json
```diff
@@ -52,6 +53,7 @@
     "playwright": "^1.51.1",
     "postcss": "^8.5.3",
     "solid-devtools": "^0.33.0",
+    "solid-transition-group": "^0.3.0",
```
incognitojam, line 56: this is a normal dep instead of a dev dep?

package.json
```diff
@@ -40,6 +40,7 @@
   },
   "devDependencies": {
     "@biomejs/biome": "1.9.4",
+    "@solid-primitives/timer": "^1.4.0",
```
incognitojam, line 43: is this used?

package.json
greatgitsby: sorry, used to putting everything in dev for FE apps. will move it since it's the standard here

package.json
greatgitsby: nice catch, was using it but not anymore. removing in next rev

src/components/upload-queue/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    .filter((i) => i.status === 'uploading').length,
+              },
+              {
+                label: 'Waiting',
+                value: () =>
+                  queue()
+                    ?.items()
+                    .filter((i) => i.status === 'queued').length,
+              },
+              { label: 'Total', value: () => queue()?.items().length },
```
greatgitsby, line 34: this formatting can be more compact but Biome was doing this.

src/hooks/use-upload-queue.ts
```diff
@@ -0,0 +1,174 @@
...
+    return allItems.sort((a, b) => {
+      const statusDiff = getStatusPriority(a.status) - getStatusPriority(b.status)
+      if (statusDiff !== 0) return statusDiff
+      const routeDiff = a.route.localeCompare(b.route)
+      if (routeDiff !== 0) return routeDiff
+      const segmentDiff = a.segment - b.segment
+      if (segmentDiff !== 0) return segmentDiff
+      return a.filename.localeCompare(b.filename)
+    })
+  })
```
greatgitsby, line 162: i think i will setup a stable sort for the uploader queue on Athena. we might be able to let the device dictate the order (more than just priority) so that the animation is consistent

for now we can at least stable sort it visually

src/components/upload-queue/QueueItem.tsx
```diff
@@ -0,0 +1,52 @@
+import { Component, createMemo } from 'solid-js'
```
incognitojam, line 1: I would just put all of the components in the same file if they aren't going to be used elsewhere in the app

src/types.d.ts
```diff
@@ -214,3 +214,15 @@ export interface UploadQueueItem {
   retry_count: number
   url: string
 }
+
+export interface UploadItem {
```
incognitojam, line 218: Can this live in an `UploadQueue.tsx` too?

src/components/upload-queue/QueueItemTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { For, Match, Switch } from 'solid-js'
+import { Transition, TransitionGroup } from 'solid-transition-group'
+
+import Icon from '~/components/material/Icon'
+import { UploadItem } from '~/types'
+
+import QueueItem from './QueueItem'
+import clsx from 'clsx'
+
+const animations = (slide: boolean) => {
```
incognitojam, line 11: Animations look like something which could just be added in a separate PR later, where there is time to test them properly

incognitojam (changes requested):
The upload queue looks good and is polished with animations and everything, but I don't think I can review this all at once still. It's just not possible for me to look at how this all works and give good feedback or ask questions...

What does a minimum upload queue look like? Could you make a version of this under 200 lines? Then you can open as many PRs as you like afterwards to add things (if you think these can be split out, just some ideas):
- animations
- clearing the queue
- showing the offline queue

By the way, this is 9% of the lines in the app!

greatgitsby:
hey, thanks for the review. sure, i can pair this down to start..

greatgitsby:
breaking these features into separate PRs won't change the lines of code it will introduce into the app. let's forgo them until there is a need, i guess.

greatgitsby:
closing in favor of https://github.com/commaai/connect/pull/251, a simplified version of this PR

PR #228, opened 2025-03-23
perf: reduce Material Symbols font payload

incognitojam:
Specify icon names to load from Material Symbols font, reducing the font payload from ~1MB to ~9KB.

Note that all icons used *must* be added to this list.

**Before**
image

**After**
image


incognitojam:
We might encounter issues with this in the future
https://github.com/google/material-design-icons/issues/1834

PR #230, opened 2025-03-23
chore: setup biome format, fix on commit

### LINKED ISSUE #220: configure biome.js style rules
optimize for LoC limits and small things like quote type, trailing comma, no semicolon, etc

incognitojam:
- `bun check`: runs `biome check`
- `bun fix`: runs `biome check --write`
- `pre-commit` hook automatically runs checks and applies fixes

closes #220

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,4 +7,5 @@ end_of_line = lf
 indent_size = 2
 indent_style = space
 insert_final_newline = true
+max_line_length = 120
 trim_trailing_whitespace = true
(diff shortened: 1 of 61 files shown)


incognitojam:
some of the wrapping, especially in import statements, isn't ideal

greatgitsby:
> some of the wrapping, especially in import statements, isn't ideal

https://biomejs.dev/reference/configuration/#formatter

can we leverage any overrides from default to get the desired behavior? default is 80char line length,

incognitojam:
`max_line_length = 120` was still wrapping too many things, `140` is better

I'm not a fan of the formatting style myself but if it's identical for everyone and happens automatically on commit then I think it'll be fine

PR #239, opened 2025-03-25
refactor: create StatisticsBar

greatgitsby:
Would love this StatisticBar for https://github.com/commaai/connect/pull/203/ to reuse spacing and styling

also inlined a few types to reduce LoC

### DIFF
--- a/src/components/RouteStatistics.tsx
+++ b/src/components/RouteStatistics.tsx
@@ -1,55 +1,30 @@
-import { createResource, Suspense } from 'solid-js'
+import { createResource } from 'solid-js'
 import type { VoidComponent } from 'solid-js'
-import clsx from 'clsx'
 
 import { TimelineStatistics, getTimelineStatistics } from '~/api/derived'
 import type { Route } from '~/types'
 import { formatDistance, formatRouteDuration } from '~/utils/format'
+import StatisticBar from './StatisticBar'
 
-const formatEngagement = (timeline?: TimelineStatistics): string => {
-  if (!timeline) return ''
+const formatEngagement = (timeline?: TimelineStatistics): string | undefined => {
+  if (!timeline) return undefined
   const { engagedDuration, duration } = timeline
   return `${(100 * (engagedDuration / duration)).toFixed(0)}%`
 }
 
-const formatUserFlags = (timeline?: TimelineStatistics): string => {
-  return timeline?.userFlags.toString() ?? ''
-}
-
-type RouteStatisticsProps = {
-  class?: string
-  route?: Route
-}
-
-const RouteStatistics: VoidComponent<RouteStatisticsProps> = (props) => {
+const RouteStatistics: VoidComponent<{ class?: string; route: Route }> = (props) => {
   const [timeline] = createResource(() => props.route, getTimelineStatistics)
 
   return (
-    <div class={clsx('flex size-full items-stretch gap-8', props.class)}>
-      <div class="flex grow flex-col justify-between">
-        <span class="text-body-sm text-on-surface-variant">Distance</span>
-        <span class="font-mono text-label-lg uppercase">{formatDistance(props.route?.length)}</span>
-      </div>
-
-      <div class="flex grow flex-col justify-between">
-        <span class="text-body-sm text-on-surface-variant">Duration</span>
-        <span class="font-mono text-label-lg uppercase">{formatRouteDuration(props.route)}</span>
-      </div>
-
-      <div class="hidden grow flex-col justify-between xs:flex">
-        <span class="text-body-sm text-on-surface-variant">Engaged</span>
-        <Suspense>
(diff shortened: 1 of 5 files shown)


src/components/RouteStatistics.tsx
```diff
@@ -1,55 +1,30 @@
...
-      </div>
-    </div>
+    <StatisticBar
+      class={props.class}
+      statistics={[
+        { label: 'Distance', value: formatDistance(props.route?.length) },
+        { label: 'Duration', value: formatRouteDuration(props.route) },
+        { label: 'Engaged', value: formatEngagement(timeline()) },
+        { label: 'User flags', value: timeline()?.userFlags.toString() },
+      ]}
```
incognitojam, line 26: I think we need to pass the values as an Accessor, otherwise we will trigger suspense and wait for all of the values to be ready

src/components/StatisticBar.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const StatisticBar: VoidComponent<{ class?: string; statistics: { label: string; value: unknown }[] }> = (props) => {
+  return (
+    <div class="flex flex-col">
+      <div class={clsx('flex h-auto w-full justify-between gap-8', props.class)}>
+        <For each={props.statistics}>
+          {(statistic) => (
+            <div class="flex basis-0 grow flex-col justify-between">
+              <span class="text-body-sm text-on-surface-variant">{statistic.label}</span>
+              <Show
```
incognitojam, line 12: Then you can change this back to a `<Suspense>`

src/components/StatisticBar.tsx
```diff
@@ -0,0 +1,26 @@
...
+                when={statistic.value !== undefined}
+                fallback={<div class="w-auto skeleton-loader bg-surface-container-low rounded-sm" />}
+              >
+                <span class="font-mono text-label-lg uppercase">{statistic.value?.toString()}</span>
```
incognitojam, line 16: I don't think you'd need to have a `.toString()` here

src/utils/format.ts
```diff
@@ -19,8 +19,8 @@ const isImperial = (): boolean => {
   return locale.startsWith('en-us') || locale.startsWith('en-gb')
 }
 
-export const formatDistance = (miles: number | undefined): string => {
-  if (miles === undefined) return ''
+export const formatDistance = (miles: number | undefined): string | undefined => {
+  if (miles === undefined) return undefined
```
incognitojam, line 23: Is there a reason to change this?

incognitojam (changes requested):

src/components/StatisticBar.tsx
greatgitsby: i tried this at first but couldn't get it working. 

this is why i changed the formatters to pass thru the undefined. i managed to get the same behavior passing "optionals" through

let me try this again.

src/components/StatisticBar.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div class="flex flex-col">
+      <div class={clsx('flex h-auto w-full justify-between gap-8', props.class)}>
+        <For each={props.statistics}>
+          {(statistic) => (
+            <div class="flex basis-0 grow flex-col justify-between">
+              <span class="text-body-sm text-on-surface-variant">{statistic.label}</span>
+              <Show
+                when={statistic.value !== undefined}
+                fallback={<div class="w-auto skeleton-loader bg-surface-container-low rounded-sm" />}
```
greatgitsby, line 14: forgot to set an explicit height for the loader, easy fix!

src/components/StatisticBar.tsx
```diff
@@ -0,0 +1,23 @@
...
+const StatisticBar: VoidComponent<{ class?: string; statistics: { label: string; value: () => unknown }[] }> = (props) => {
+  return (
+    <div class="flex flex-col">
+      <div class={clsx('flex h-auto w-full justify-between gap-8', props.class)}>
+        <For each={props.statistics}>
+          {(statistic) => (
+            <div class="flex basis-0 grow flex-col justify-between">
+              <span class="text-body-sm text-on-surface-variant">{statistic.label}</span>
+              <Suspense fallback={<div class="h-[20px] w-auto skeleton-loader rounded-xs" />}>
+                <span class="font-mono text-label-lg uppercase">{statistic.value()?.toString() ?? '—'}</span>
```
greatgitsby, line 13: The fallback to `undefined` for the formatters is so that we can gracefully display `—`. We already do this for other format functions (like route duration)

incognitojam: Sorry where do we do that?

incognitojam: oh I just read this line wrong because it went off the screen :sweat_smile: 

image

looks good!

incognitojam (approved):
Looks good!

PR #243, opened 2025-03-25
cleanup(dashboard): fetch devices where they are used

incognitojam:
Instead of prop drilling, we can just fetch the list of devices where they are used. We should use a query library to cache/de-duplicate requests.


greatgitsby:
i'll make an issue to refactor upload queue once we onboard that request lib. it will simplify the polling logic that is required.

Issue #222, opened 2025-03-22
scrub through video using map


BBBmau:
similar to https://github.com/commaai/connect/issues/64 only difference is we use the car marker as a way to scrub through the timeline also. Not sure if it makes sense to keep the ability to scrub through timeline when this gets supported but would be nice to add in and figure that out afterwards

- [ ] include car location in map, it only shows the route and not where the car is throughout the route. [implementation in old connect](https://github.com/commaai/connect-pwa-archive/blob/9fd6bff142b60a155e85b457e187ebf13b1c7650/src/components/DriveMap/index.jsx#L143-L151)
- [ ] implement scrubbing with map

PR #255, opened 2025-03-27
Interactive route map with real-time progress

### LINKED ISSUE #213: Show real-time replay location in map
Old connect would show the location of the vehicle during replays and move the map to follow the location.

Image

Currently in the demo it just shows the entire route without moving or any indication of where the video is happening at. This makes it very difficult to find a specific event at any location.

Image

It would also be great to be able to select a point along the route to see the video at that specific time.

Cynamide:
Resolves #213 and #212 

I know someone else is already working on it but thought I should try it anyways. Let me know if the map is intuitive to use. 
Also, I bumped up the location of map just below the video so its easier to use with the video playing. The map is by default static for easy scrolling. Users can click on any path on the map to jump to that location. Also drag the marker for finer control. Both these cases will make the map interactive. A small icon appears on the top left which will make the map static again.

image


sshane:
The dragging of the car is pretty cool! Can we smoothly interpolate the position like old connect does?

https://github.com/user-attachments/assets/b49b9d2c-c4d8-4f67-be4e-64ae23929371

Cynamide:
Thank you so much @incognitojam

Cynamide:
Right now, I am not interpolating the points and fully relying on the marker CSS to do the work for me with way less code. Let me know if interpolating looks necessary ( when the map is fully zoomed in, the marker is small enough to be between points where you can make out the grey and purple lines distinctly updating before marker moves). 

Also found a little bug where map doesn't reload when a different route is selected. Will work on that.

TheSecurityDev:
I like how the map is closer to the video, and how the route is darker for the past history. I may have to steal that for my PR #224. I also have the same issue with interpolation, but it's something that could be implemented if necessary, it just makes it a bit more complex. My PR is already a bit more complex due to the logic for locking the map while scrolling.

Cynamide:
Thats right, even after adding interpolation, there is still a need for smooth css transitions which is why I removed interpolation completely. 
You are welcome to use my code for your PR. Also, yeah I'm working on adding cases where we would not need the CSS/panning transitions.

Cynamide:
The reason I'm keeping a bigger icon is so that you cant see the purple and grey paths updating their positions behind the icon. also locked the max zoom-in for the map so that the icon doesn't become small enough to see this update.

TheSecurityDev:
> The reason I'm keeping a bigger icon is so that you cant see the purple and grey paths updating their positions behind the icon. also locked the max zoom-in for the map so that the icon doesn't become small enough to see this update. Looks like I'm all done now.

Can't you have the division always right in the center and follow the icon smoothly?

TheSecurityDev:
Also kind of curious, I just noticed you referenced pull request 212, which doesn't really seem relevant. Perhaps you meant #222?

Cynamide:
> > The reason I'm keeping a bigger icon is so that you cant see the purple and grey paths updating their positions behind the icon. also locked the max zoom-in for the map so that the icon doesn't become small enough to see this update. Looks like I'm all done now.
> 
> Can't you have the division always right in the center and follow the icon smoothly?

I could, but the problem is the polyline is updated instantly instead of having an animation like the marker. I'd have to calculate the points separately each time. I'll see if leaflet allows for animating polylines too. That should really fix it.

Cynamide:
> Also kind of curious, I just noticed you referenced pull request 212, which doesn't really seem relevant. Perhaps you meant #222?

Oh right my bad. Thanks for pointing it out!

TheSecurityDev:
> > Can't you have the division always right in the center and follow the icon smoothly?
> 
> I could, but the problem is the polyline is updated instantly instead of having an animation like the marker. I'd have to calculate the points separately each time. I'll see if leaflet allows for animating polylines too. That should really fix it.

It just seems like too much of a hack, and it's too bad to have to reduce the available zoom.

Cynamide:
> > > Can't you have the division always right in the center and follow the icon smoothly?
> > 
> > 
> > I could, but the problem is the polyline is updated instantly instead of having an animation like the marker. I'd have to calculate the points separately each time. I'll see if leaflet allows for animating polylines too. That should really fix it.
> 
> It just seems like too much of a hack, and it's too bad to have to reduce the available zoom. I would almost rather have the visual bug.
> 
> Edit: I guess the zoom is far enough to be usable, so maybe not a big deal, but I would like to have the icon smaller still. Again, not a big deal.

You are right about the icon size. Let me see what I can do.

TheSecurityDev:
I really like the mechanics of this compared to mine in many ways, but I have a few more suggestions.

1. Enable the zoom control buttons on the top or bottom right so we can at least zoom without having to drag along the route to unlock, since it might not be immediately obvious how to do it.
2. Add some text to the center/lock button, or else adjust the icon so it's centered properly (there's a tiny bit of space on the right).
3. It seems like clicking anywhere in the map should unlock it, but I understand how hard it is to implement that properly (which is why my PR ended up more complicated). It's something to think about, but it's probably fine the way it is.
4. It would be nice if it would pan the map as you are dragging if you drag out of the map.
5. In light that dragging/selecting on the route is required to enable panning/zooming, there's no way to drag clear to the end without auto-panning the map like I suggested, so I think the map should be initially zoomed out far enough so the entire route can be displayed on the map before it gets centered. This way it would be easier to initially activate the map and drag to the part of the route you want in a single action.

Cynamide:
> I really like the mechanics of this compared to mine in some ways, but I have a few more ideas/suggestions.
> 
> 1. Enable the zoom control buttons on the top or bottom right so we can at least zoom without having to drag along the route to unlock, since it might not be immediately obvious how to do it.
> 2. Add some text to the center/lock button, or else adjust the icon so it's centered properly (there's a tiny bit of space on the right).
> 3. It seems like clicking anywhere in the map should unlock it, but I understand how hard it is to implement that properly (which is why my PR ended up more complicated). It's something to think about, but it's probably fine the way it is.
> 4. It would be nice if it would pan the map as you are dragging if you drag out of the map.
> 5. In light that dragging/selecting on the route is required to enable panning/zooming, there's no way to drag clear to the end without auto-panning the map like I suggested, so I think the map should be initially zoomed out far enough so the entire route can be displayed on the map before it gets centered. This way it would be easier to initially activate the map and drag to the part of the route you want in a single action.
> 6. If possible, make it pan quicker since it takes a bit to catch up when seeking via the timeline.

1. That's something that I was debating to add or not. I now feel like we should have a more stricter map control so that there are no visual bugs while zooming in or out ( like pinching the map on mobile phones and the frustration you described). Zooming if done only with the  buttons would be more deterministic (and good). 
2. Oh nice catch. I'll center it. There is also another way on the old connect that uses debounce to lock in after a couple secs of inactivity. doing that would remove the button which could make the map cleaner. Not sure whats more intuitive to use.
3. I purposely didn't add that since on mobile phones, it'll be difficult to scroll since most of the screen has the video, map, timeline (all interactable).  
4. Like the debounce I mentioned earlier?
5. Sounds good. Though I'm still not sure how we would tell the user that they can select/drag to unlock the map. An overlay for a few seconds sounds too much. Or maybe like a popup on the map that stays for a few seconds should do it.
6. Good catch, I'll fix that.

Cynamide:
Added a popup that shows how to interact with the map with a hide option. lmk how it looks. all other panning/transition bugs are fixed afaik.

TheSecurityDev:
The icon looks squished now. Here's the code I used for mine to give you an idea:
```ts
const createMarkerIcon = () => {
    const el = document.createElement('div')
    render(
      () => (
        <div class="flex size-[30px] items-center justify-center rounded-full bg-primary-container">
          <Icon name="directions_car" size="20" />
        </div>
      ),
      el,
    )

    return Leaflet.divIcon({
      className: 'border-none bg-none',
      html: el.innerHTML,
      iconSize: [20, 20],
      iconAnchor: [15, 15],
    })
  }
```

TheSecurityDev:
The message at the bottom is cool, but I'm not sure it should be visible by default every time. Maybe a help icon or something, but it should be intuitive enough it's not necessary.

TheSecurityDev:
Also I'm not sure about the lock icons. I think it almost makes more sense to use 'my_location', 'location_searching' icons respectively, but that's more to clarify the auto tracking part. The lock icons are more helpful to clarify whether the map controls are locked or not. Just something to think about.

TheSecurityDev:
Also one minor thing: Is it possible to make it move to the route position on click down instead of when you release the button? This would make it so you could click and drag along a route anywhere without having to drag the marker specifically. Right now, if you click and drag a random part of the route it just drags the map when it feels like it should put the marker there and drag along with the cursor.

TheSecurityDev:
> * Like the debounce I mentioned earlier?

No I'm talking about when you drag the marker outside of the visible map boundary, it should automatically pan the map so you can keep dragging. Otherwise you have to zoom out to see enough of the route, which is another reason I want to keep the zoom controls on the scroll wheel as well as the buttons if possible.

TheSecurityDev:
I created a branch: https://github.com/TheSecurityDev/comma-connect/tree/cynamide-interactive-map

I fixed the icon size there. If you add my repo as another origin to your project, you should be able to take from my branch and merge into yours and I can do the same. This way we can both contribute.

TheSecurityDev:
> All very useful feedbacks. Thanks! I'm afraid I'm not that great with github and so I dont want to mess up my local/origin branch in any way. I applied some of your suggestions. Could you take a look and let me know if it looks good? I haven't added zoom gestures because that would introduce visual glitches while doing the gestures. Other than that lmk!

You should be able to just add a new remote in VSCode and then fetch it to view the branches, then merge my branch into yours. If there are merge conflicts you will have to solve those, but there shouldn't be unless you've also made changes. It would be nice, but oh well. I guess I won't bother doing too much with it then.

But I really wish you could fix the zoom controls. I'm not sure what you mean by visual glitches, can you explain? I implemented it successfully in my branch.

I see also dragging is currently broken. It seems it's always centering even when it's not locked.

Also, the latest master just had some changes merged that combines the timeline with the video player, so can you try to fix that so the timeline is just below the player (but connected). Check the latest master preview to see what I mean.

TheSecurityDev:
You can always reset your branch by doing `git reset --hard origin/interactive-map`. Also another tip is try to only do a single thing per commit. That way you can revert by just reverting that commit. I would just undo panning the map when handling the drag, since that's causing the issue it seems. We can always do other things later, but we should get it into a presentable state now at least and fix all the bugs.

TheSecurityDev:
> I added the centering while dragging so that you could see a little ahead while dragging. But I realize like you pointed it out that its really not that much more intuitive. I'll try merging your branch.

I sent you a message on Discord, if you have any issues we could discuss there too.

Cynamide:
> I sent you a message on Discord, if you have any issues we could discuss there too.

Hmm....I don't see any message

TheSecurityDev:
> > I sent you a message on Discord, if you have any issues we could discuss there too.
> 
> Hmm....I don't see any message

Discord hides them more now to prevent spam, so you might have to find the message requests section or something. Check the pending tab.

Also I just merged in some updates from master to fix the timeline and the video playing.

TheSecurityDev:
I guess another thing we could do is I could open a pull request on your repo and you could merge them that way if you'd like. I didn't know that was possible until now.

PR #257, opened 2025-03-27
Fix pixelated flag icon border for certain screens

ugtthis:
- was using `ring-1` changed to `border`
- seems like it gets jaggedy and pixelated on certain screens
- using `border` better/consistent for lower DPI screens

From Shane:
pixelated-flag


incognitojam:
I couldn't repro the original issue

sshane:
It looks better, but it still looks out of place IMO. It's the only brightly colored icon in the UI

ugtthis:
CleanShot 2025-03-27 at 15 45 00@2x

Currently there isn't a key on what the different colors mean on the timeline. I matched the icon color to the timeline flag color to implicitly tell users that they are related. If we changed the color of the flag to a non yellow color it would be more inconsistent to the timeline colors

PR #271, opened 2025-03-27
live.sh setup and dev script

sshane:


live.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd $DIR
+
+if [ ! -d "$HOME/.bun" ]; then
+  curl -fsSL https://bun.sh/install | bash
```
incognitojam, line 8: need to add bun to path if it's first install

PR #272, opened 2025-03-28
feat: poll offline queue

greatgitsby:
https://github.com/commaai/connect/issues/51

tested locally:
1. when device offline, upload request puts items in offline queue, showing them in the queue in the UI with a "Queued" message
2. when device comes online, status switches to "Waiting" until Athena server pushes the queued requests
3. finally, messages start coming in and processing as normal


incognitojam:
Is it expected for it to just say "Waiting for device to connect..." when the device is offline still?

greatgitsby:
that was the intention but i'm open to another approach.

greatgitsby:
"waiting to connect" indicates that device is offline or coming online

incognitojam:
I think if we know the device is offline (and I guess, if the offline queue is empty) we should say "device offline"

incognitojam (approved):

Issue #244, opened 2025-03-26
Text overflows on smaller screens


ugtthis:
- Location text overflows when smaller screen
- Time text "AM" and year end up on a new line, maybe better if that text sizes down so it stays on one line

![Image](https://github.com/user-attachments/assets/8e415f83-395c-4fb2-b697-4e44031320ff)

sshane:
Is this because position is absolute or something? Let's just push down elements when this happens

PR #273, opened 2025-03-28
Clearer text layout on card

ugtthis:
- clearer text hierarchy using separate lines for different info
- better adapts to smaller screens like iphone SE
- resolves issue #244 
- rm unused sub components in `Card.tsx`
- if location text is long it goes to the next line
- location display logic - when only one location on card it shows "Stayed near:"(example at bottom). I can change this to something else if there is a better label that suggests your trip movement was very minimal

## iPhone SE screen width
| Old UI | New UI |
|--------|--------|
| Old-ui | New-ui |

<br>

---

<br>

## "Stayed Near:" if only 1 location shows 
ifSingle-ui


greatgitsby:
"Stayed near" reads odd to me... what are your thoughts about simply "Near"?

incognitojam:
I like the date/time change but not sure about adding "from" and "to"..

sshane:
The card height is huge now

image

sshane:
Can we just prefer to copy current connect's UI for most things? It's two separate projects to port its functionality *and* come up with new design

ugtthis:
Closing since PR #356 will make it easier to read the cards without taking up to much screen real estate for now. Will fix #244, and rm subcomponents in seperate PRs

PR #289, opened 2025-03-28
Remove duplicate events.json queries

sshane:
It was querying the files twice for every segment for every route

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -25,7 +25,7 @@ jobs:
       - run: bun install --frozen-lockfile
       - run: bun playwright install
         if: steps.playwright-cache.outputs.cache-hit != 'true'
-      - run: bun check
+      - run: bun check --reporter=github
       - run: bun run test
       - run: bun lines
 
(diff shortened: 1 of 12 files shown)


sshane:
We need to make sure we're not doing anything dumb like this, are there any simple tests we can implement for duplicate queries?

Or maybe it's better to bake everything into the Route class that's passed around so this is impossible @incognitojam

incognitojam:
Most apps use a query client to centralise deduplicating requests, caching results, refetching on an internal etc.

I think @greatgitsby is looking at adding TanStack Query which is basically the standard lib for this

incognitojam:
When you `bun install`-ed it should have setup a commit hook to run the formatter

sshane:
PyCharm doesn't support it so I uninstalled it :(

sshane:
Wait, where is it?

incognitojam (approved):

incognitojam:
> PyCharm doesn't support it so I uninstalled it :(

you should be able to get a biome plugin too, can format on save

PR #298, opened 2025-03-29
Remove unnecessary routes_segments query

sshane:
Added in https://github.com/commaai/connect/pull/216

from `routes_segments` API:

```
    for row in q.all():
      d["start_time_utc_millis"] = int((0 if r.start_time is None else r.start_time.timestamp()) * 1000)
      d["end_time_utc_millis"] = int((0 if r.end_time is None else r.end_time.timestamp()) * 1000)

      # TODO: remove these once connect doesn't use them
      d["segment_numbers"] = list(range(0, r.maxqlog+1))
      d["segment_start_times"] = [d["start_time_utc_millis"]+(i*60*1000) for i in range(r.maxqlog+1)]
      d["segment_end_times"] = [min(d["segment_start_times"][i]+(60*1000), d["end_time_utc_millis"]) for i in range(r.maxqlog+1)]

      res[r.fullname] = d

  return jsonify(list(res.values()))
```

### DIFF
--- a/src/api/derived.ts
+++ b/src/api/derived.ts
@@ -77,6 +77,7 @@ export interface TimelineStatistics {
 }
 
 const getDerived = <T>(route: Route, fn: string): Promise<T[]> => {
+  console.log(route.fullname, fn)
   let urls: string[] = []
   if (route) {
     const segmentNumbers = Array.from({ length: route.maxqlog }, (_, i) => i)
--- a/src/api/derived.ts
+++ b/src/api/derived.ts
@@ -77,7 +77,6 @@ export interface TimelineStatistics {
 }
 
 const getDerived = <T>(route: Route, fn: string): Promise<T[]> => {
-  console.log(route.fullname, fn)
   let urls: string[] = []
   if (route) {
     const segmentNumbers = Array.from({ length: route.maxqlog }, (_, i) => i)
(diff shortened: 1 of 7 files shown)


src/components/RouteActions.tsx
```diff
@@ -35,7 +35,7 @@ interface RouteActionsProps {
 }
 
 const RouteActions: VoidComponent<RouteActionsProps> = (props) => {
-  const [routeResource] = createResource(() => props.routeName, getRoute)
+  const [routeResource] = createResource(() => props.routeName, getRoute)  // wtf with the duplicate queries, why not pass route in?
```
sshane, line 38: @incognitojam do you know why we sometimes pass route and other times pass routeName for the child component (?) to re-query the route when the parent had it? Seems confusing

src/api/derived.ts
```diff
@@ -77,6 +77,7 @@ export interface TimelineStatistics {
 }
 
 const getDerived = <T>(route: Route, fn: string): Promise<T[]> => {
+  console.log(route.fullname, fn)
```
sshane, line 80:
```suggestion
```

sshane:
@incognitojam it's very important to take extra time to do the simplest thing

PR #299, opened 2025-03-29
feat: use maxqlog as proxy for segment count

greatgitsby:


greatgitsby:
must be a flaky test, not sure how this failure relates to the PR

src/components/RouteUploadButtons.tsx
```diff
@@ -109,7 +109,7 @@ const RouteUploadButtons: VoidComponent<RouteUploadButtonsProps> = (props) => {
     const fileTypesToUpload = buttonToFileTypeMap[type]
 
     try {
-      await uploadAllSegments(props.routeName, route.segment_numbers.length, fileTypesToUpload)
+      await uploadAllSegments(props.routeName, route.maxqlog, fileTypesToUpload)
```
sshane, line 112: The original code adds 1 since maxqlog is an index, not count: `d["segment_numbers"] = list(range(0, r.maxqlog+1))`

Already opened a PR here: https://github.com/commaai/connect/pull/298

sshane: This also doesn't address the point of removing the segments endpoint we want to switch away from for timeless routes support

PR #302, opened 2025-03-29
Remove 1/2 duplicate route query

sshane:
Ideally we only make one /route query when you load a route, not three. This PR removes one.

https://github.com/commaai/connect/pull/298#discussion_r2019681192


src/api/index.ts
```diff
@@ -9,6 +9,7 @@ const populateFetchedAt = <T>(item: T): T => {
 }
 
 export async function fetcher<T>(endpoint: string, init?: RequestInit, apiUrl: string = BASE_URL): Promise<T> {
+  console.log("fetcher", endpoint)
```
sshane, line 12:
```suggestion
```

PR #303, opened 2025-03-29
Remove 2/2 duplicate route queries

sshane:
pt. 2 of https://github.com/commaai/connect/pull/302

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -9,6 +9,7 @@ const populateFetchedAt = <T>(item: T): T => {
 }
 
 export async function fetcher<T>(endpoint: string, init?: RequestInit, apiUrl: string = BASE_URL): Promise<T> {
+  console.log('fetcher', endpoint)
   const res = await fetch(`${apiUrl}${endpoint}`, {
     ...init,
     headers: {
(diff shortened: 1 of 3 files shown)


src/components/Timeline.tsx
```diff
@@ -92,16 +92,17 @@ function renderTimelineEvents(route: Route | undefined, events: TimelineEvent[])
 interface TimelineProps {
   class?: string
   routeName: string
+  route: Accessor<Route | undefined>
```
sshane, line 95: @incognitojam I couldn't figure out how to pass the non-accessor as it was always undefined, but maybe we want to pass the accessor by default? Have any preference?

sshane: Edit: I could pass it, but the loading animation breaks

src/api/index.ts
```diff
@@ -9,6 +9,7 @@ const populateFetchedAt = <T>(item: T): T => {
 }
 
 export async function fetcher<T>(endpoint: string, init?: RequestInit, apiUrl: string = BASE_URL): Promise<T> {
+  console.log('fetcher', endpoint)
```
sshane, line 12:
```suggestion
```

PR #310, opened 2025-03-29
Route list: group by day

sshane:


sshane:
I couldn't get the reactivity to work without the entire page refreshing every time it updated the routes, or duplicate headers. I think this is a good compromise, but let me know if you have any suggestions @incognitojam

incognitojam:
Might try re-writing the route list query so that we don't have to render it in chunks

sshane:
I couldn't get it to render from a flat array without the entire page redrawing when you add a new route.

greatgitsby:
try using a Solid.js store with reconcile(). it only updates the diff

sshane:
I think the result of a group-like function was the part that it couldn't diff properly, so will need to store the routes grouped which requires a tiny re-write

incognitojam:
> I couldn't get it to render from a flat array without the entire page redrawing when you add a new route.

It was because there was a resource in RouteCard without a suspense wrapping it, fixed in #400

sshane:
You can revert and re-do with a pre-grouped object if you can get it working

PR #311, opened 2025-03-29
Fix routes not loading because one is missing events.json

sshane:
If there was an uncaught error, the Suspense would never get triggered and be stuck in fallback state forever

before:

image

after:

image


sshane:
This is still broken:

image

sshane:
@incognitojam can we add a CI test to catch missing exception handling for any API call?

PR #314, opened 2025-03-29
Revert "Remove duplicate events.json queries (#289)"

sshane:
This reverts commit fcd306d92ae87cfa9a542a359d041b15d428459b.

There was another instance of RouteStatistics that didn't set timeline.

### DIFF
--- a/src/components/RouteStatistics.tsx
+++ b/src/components/RouteStatistics.tsx
@@ -1,6 +1,7 @@
+import { createResource } from 'solid-js'
 import type { VoidComponent } from 'solid-js'
 
-import { TimelineStatistics } from '~/api/derived'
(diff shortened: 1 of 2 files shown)


sshane:
@incognitojam let's figure out some clean way to ensure we only query for route files once

greatgitsby:
once i finish setup for TanStack Query, we'll get this for free as we migrate queries

https://github.com/commaai/connect/pull/326

incognitojam:
The browser actually caches these requests for free, although I guess there is overhead in making the requests again for every segment in a route, and then processing the data again.

greatgitsby:
correct; we don't have to download the file, however we still have to make the request to Azure, browser looks at cache header, etc

PR #326, opened 2025-03-29
feat: setup `@tanstack/solid-query` for data fetching, use in upload queue

greatgitsby:
https://tkdodo.eu/blog/react-query-data-transformations
https://tkdodo.eu/blog/testing-react-query
https://tkdodo.eu/blog/the-query-options-api#query-factories

- setup `solid-query` for app, dev tools integration (**note**: [does not bundle in prod](https://tanstack.com/query/latest/docs/framework/solid/devtools)). see PR deployment
- ported existing behavior to `solid-query`
  - online queue is active, refetching on interval. offline queue always polling as well
  - once the queue has loaded at least once, the queue holds its current state. when the drawer is opened later, it will present that last known state and fetch the latest (e.g. you were uploading 10 files with the queue open, closed it, and reopened it. once you reopen, it'll show the progress from the last known fetch and schedule it to be updated). UX should be smoother (this is the normal behavior of `solid-query`)
- follow up PRs can begin to migrate usage to `createQuery`

test cases:
- [X] on load, "waiting for device" msg
- [X] device connected, nothing uploading, "nothing to upload" msg
- [X] device connected, upload file, files appear in queue
- [X] device disconnected, upload file, appears in offline queue; on connect, eventually[^1] messages appear
- [X] device connected, uploading, reboot device; files go away, device in offline state; once reconnected, switches back to uploading
- [X] device disconnected, nothing in offline queue, "device offline" msg
- [X] device disconnected, something in offline queue, shows offline messages
- [X] cancel item
- [X] cancel all
- [x] uploading files, switch to new device; show that device's upload queue state

[^1]: the transition between offline and online with offline messages present is undefined behavior (could say "device offline" then messages appear, or "nothing to upload" then appear). existing behavior, created issue: https://github.com/commaai/connect/issues/354


src/index.tsx
```diff
@@ -17,4 +19,12 @@ const root = document.getElementById('root')
 
 if (!root) throw new Error('No #root element found in the DOM.')
 
-render(() => <App />, root)
+render(
+  () => (
+    <QueryClientProvider client={new QueryClient()}>
```
incognitojam, line 24: Things in this file don't have any test coverage at the moment because we just import `<Routes />` from `App.tsx` in `App.browser.test.tsx`...

I think if `UploadQueue` was visible on page load it would have caused the test to fail from trying to access the query client?

tl;dr move these to `AppLayout` in `App.tsx` (I just created a PR to add it in #358)

src/components/UploadQueue.tsx
```diff
@@ -61,76 +54,63 @@ const UploadQueueRow: VoidComponent<{ dongleId: string; item: DecoratedUploadQue
...
+  <div class="flex items-center gap-2">
+    <Icon name={props.icon} class={props.iconClass} />
+    <span class="text-body-lg">{props.message}</span>
+  </div>
+)
 
 const UploadQueue: VoidComponent<{ dongleId: string }> = (props) => {
-  const [error, setError] = createSignal<string | undefined>(WAITING)
-  const [items, setItems] = createStore<DecoratedUploadQueueItem[]>([])
+  const onlineQueueKey = createMemo(() => ['online_queue', props.dongleId])
```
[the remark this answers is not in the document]
greatgitsby: i suppose it doesn't really need to as it's not a heavy computation

i'll move it to something simpler

incognitojam:
The test seems fail when `<SolidQueryDevtools />` is included..

incognitojam:
You can put `<QueryClientProvider>` and `<SolidQueryDevtools>` in `<App>` instead, and we'll create a separate wrapper and query client in tests once it's needed

greatgitsby:
great article you attached! we need to revisit this sooner than later

greatgitsby:
took the creator's advice and extracted the query defaults to make the tests more flexible

src/utils/query-client.ts
```diff
@@ -0,0 +1,14 @@
+import { QueryClient } from '@tanstack/solid-query'
+
+export const ONLINE_QUEUE = 'online_queue'
+export const OFFLINE_QUEUE = 'offline_queue'
+
+export const getAppQueryClient = () => {
+  const queryClient = new QueryClient()
+
+  const pollingConfig = { retry: false, refetchInterval: 1000 }
+  queryClient.setQueryDefaults([ONLINE_QUEUE], pollingConfig)
```
incognitojam, line 10: Interesting that this is recommended - moving it away from the component feels bad. Although I guess defining all of the keys somewhere is nice?

src/components/UploadQueue.tsx
```diff
@@ -61,76 +55,59 @@ const UploadQueueRow: VoidComponent<{ dongleId: string; item: DecoratedUploadQue
...
+)
 
 const UploadQueue: VoidComponent<{ dongleId: string }> = (props) => {
-  const [error, setError] = createSignal<string | undefined>(WAITING)
-  const [items, setItems] = createStore<DecoratedUploadQueueItem[]>([])
+  const onlineQueueKey = () => [ONLINE_QUEUE, props.dongleId]
+  const onlineQueue = createQuery(() => ({
+    queryKey: onlineQueueKey(),
+    queryFn: () => getUploadQueue(props.dongleId),
+    select: (data) => data.result?.map((item) => ({ ...item, ...parseUploadPath(item.url) })).sort((a, b) => b.progress - a.progress) || [],
```
incognitojam, line 70: I think it's more performant to move the `select` impl out of `createQuery`
image

src/utils/query-client.ts
greatgitsby: you can't override them in tests (e.g. turn off polling, retry count, etc)

incognitojam: I see why it's useful for tests though!

src/components/UploadQueue.tsx
greatgitsby: i need to read his entire site before we merge this PR

greatgitsby:
i don't like the `query-client.ts` approach. we should instead migrate over the `src/api/` files to hold the queries, selects, etc for each relevant call

marking as draft for now

src/utils/query-client.ts
greatgitsby: i've removed the key definition and co-located it with the queries themselves -- i like how TkDodo proposes to set it up

greatgitsby:
moved to co-locating queries and keys together. as we migrate queries over, we can add onto the new objects.

i tested out pre-fetching the upload queue even if it's not visible -- takes 3 lines of code now 😊

src/api/athena.ts
```diff
@@ -11,20 +12,36 @@ import {
...
 
 // Higher number is lower priority
 export const COMMA_CONNECT_PRIORITY = 1
 
 // Uploads expire after 1 week if device remains offline
 const EXPIRES_IN_SECONDS = 60 * 60 * 24 * 7
 
-export const cancelUpload = (dongleId: string, ids: string[]) =>
-  makeAthenaCall<CancelUploadRequest, CancelUploadResponse>(dongleId, 'cancelUpload', { upload_id: ids })
+export const athena = {
```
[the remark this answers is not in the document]
incognitojam: I think if we want to re-use these in multiple places we can create hook-like methods like `useDevice` which wraps the API calls in a query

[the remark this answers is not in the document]
greatgitsby: makes sense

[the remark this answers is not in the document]
greatgitsby: hm, we will definitely use the query keys in at least two spots for the queues (to set query client defaults)

i'll use that "hook" pattern

greatgitsby:
@incognitojam do you think the behavior of showing last upload queue state is desirable? to me it's one less loading state which makes it feel snappier, but i can also remove this behavior

src/queries/upload-queue.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export const uploadQueue = {
+  prefix: ['upload_queue'],
+
+  online: () => [...uploadQueue.prefix, 'online'],
+  onlineForDongle: (dongleId: string) => [...uploadQueue.online(), dongleId],
+  getOnline: (dongleId: string) =>
+    queryOptions({
+      queryKey: uploadQueue.onlineForDongle(dongleId),
+      queryFn: () => makeAthenaCall<void, UploadQueueItem[]>(dongleId, 'listUploadQueue'),
```
incognitojam, line 34: I think the athena calls should stay in `src/api`

src/types.d.ts
```diff
@@ -207,4 +207,8 @@ export interface UploadQueueItem {
   progress: number
   retry_count: number
   url: string
+  route: string
```
incognitojam, line 210: This file is meant to be for API types

incognitojam (changes requested):
All of the upload queue logic should stay in the UploadQueue component

incognitojam:
> do you think the behavior of showing last upload queue state is desirable? to me it's one less loading state which makes it feel snappier, but i can also remove this behavior

That sounds fine, can it cause problems?

greatgitsby:
> That sounds fine, can it cause problems?

It should not. It will optimistically show the latest data in the query cache and make a fresh network call at the same time.

src/queries/upload-queue.tsx
incognitojam: Things like the athena methods available and return types will be in the API though, so you'd still want to export a method like `listUploadQueue` from the API package and use that here

src/queries/upload-queue.tsx
```diff
@@ -0,0 +1,23 @@
+import { createMutation, queryOptions, useQueryClient } from '@tanstack/solid-query'
+import { getUploadQueue, makeAthenaCall } from '~/api/athena'
+import { getAthenaOfflineQueue } from '~/api/devices'
+
+export const uploadQueue = {
```
incognitojam, line 5: Why can't this be in UploadQueue.tsx too?

src/api/athena.ts
```diff
@@ -18,13 +16,8 @@ export const COMMA_CONNECT_PRIORITY = 1
 // Uploads expire after 1 week if device remains offline
 const EXPIRES_IN_SECONDS = 60 * 60 * 24 * 7
 
-export const cancelUpload = (dongleId: string, ids: string[]) =>
-  makeAthenaCall<CancelUploadRequest, CancelUploadResponse>(dongleId, 'cancelUpload', { upload_id: ids })
-
 export const getNetworkMetered = (dongleId: string) => makeAthenaCall<void, boolean>(dongleId, 'getNetworkMetered')
-
 export const getUploadQueue = (dongleId: string) => makeAthenaCall<void, UploadQueueItem[]>(dongleId, 'listUploadQueue')
-
```
incognitojam, line 27: revert this

incognitojam (changes requested):

src/components/UploadQueue.tsx
```diff
@@ -1,39 +1,66 @@
...
 
-const cancel = (dongleId: string, ids: string[]) => {
-  if (ids.length === 0) return
-  cancelUpload(dongleId, ids).catch((error) => {
-    console.error('Error canceling uploads', error)
-  })
-}
-
-const UploadQueueRow: VoidComponent<{ dongleId: string; item: DecoratedUploadQueueItem }> = ({ dongleId, item }) => {
+const UploadQueueRow: VoidComponent<{ cancel: (ids: string[]) => void; item: UploadQueueItemWithAttributes }> = ({ cancel, item }) => {
```
incognitojam, line 63: You should avoid destructuring props in solid

greatgitsby: would you prefer creating a derived value like this or referencing props.item everywhere?

`const item = () => props.item`

incognitojam: that looks fine

src/components/UploadQueue.tsx
```diff
@@ -1,136 +1,119 @@
...
-        setError(error.toString())
-      })
-      .finally(() => {
-        if (!timeout) return
-        timeout = setTimeout(fetch, 1000)
-      })
-  }
+  const onlineQueue = createQuery(() => queries.getOnline(props.dongleId))
+  const offlineQueue = createQuery(() => queries.getOffline(props.dongleId))
+  const cancel = queries.cancelUpload(props.dongleId)
```
incognitojam, line 100: I think this might not be reactive - if the dongleId passed as a prop changes, this wouldn't update

Can't repro right now because we recreate the dashboard when the route changes

greatgitsby: you're right -- i removed the keyed value at the top level and it isn't reactive

i'll resolve in a bit

greatgitsby:
do we want to recreate the mutation for every row? we only need to recreate it when the dongle changes

incognitojam:
oh I see

src/App.tsx
```diff
@@ -38,10 +42,15 @@ export const AppLayout: ParentComponent = (props) => {
...
 }
 
+const queryClient = getAppQueryClient()
+
 const App: VoidComponent = () => (
-  <Router root={AppLayout}>
-    <Routes />
-  </Router>
+  <QueryClientProvider client={queryClient}>
+    <SolidQueryDevtools />
```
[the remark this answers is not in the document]
greatgitsby: no. see first point in PR description

just checked again. i don't see the tanstack dev tools in prod
image

PR #329, opened 2025-03-29
perf: make switching route snappy

incognitojam:
Avoids recreating the RouteActivity when changing between selected routes


incognitojam:
this may highlight some new bugs

incognitojam:
the RouteActivity is always destroyed when you go back to the route list on mobile

sshane:
It still seems to be slow on desktop

PR #334, opened 2025-03-29
ci: check TypeScript type errors

greatgitsby:
- run `tsc --noEmit` as a "type-checker" in the test suite (won't block the build if building works, so we can preview even if types aren't perfect)
- fix the two complaints

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -2,6 +2,7 @@ set -e
 
 bun install --frozen-lockfile
 bun biome ci
+bun tsc --noEmit
 bun playwright install
 bun run test run
 bun lines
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -1,6 +1,7 @@
 set -e
 
 bun install --frozen-lockfile
+bun tsc --noEmit
 bun biome ci
```
incognitojam, line 5:
```suggestion
bun biome ci
bun tsc
```
I think biome should go first since it's so fast, and secondly we shouldn't need `--noEmit` as we already have that set in `tsconfig.json`

greatgitsby: oh missed that, thanks

PR #338, opened 2025-03-29
fix: reset upload btn state if user changes routes

greatgitsby:
resolves: https://github.com/commaai/connect/issues/337

adds an abort signal and reset handling

we should make this stateless, i don't want to add side effects.

this component will need a refactor once we add more behavior like in https://github.com/commaai/connect/issues/258

https://github.com/user-attachments/assets/0c12af85-b276-4b69-962c-b2e62c5c90b7

### DIFF
--- a/src/components/RouteUploadButtons.tsx
+++ b/src/components/RouteUploadButtons.tsx
@@ -4,11 +4,21 @@ import clsx from 'clsx'
 
 import Icon, { type IconName } from '~/components/material/Icon'
 import Button from './material/Button'
-import { FileTypes, uploadAllSegments } from '~/api/upload'
+import { uploadAllSegments, type FileType } from '~/api/upload'
 import type { Route } from '~/types'
 
+type ButtonType = 'road' | 'driver' | 'logs' | 'route'
+type ButtonState = 'idle' | 'loading' | 'success' | 'error'
+
+const BUTTON_TO_FILE_TYPES: Record<ButtonType, FileType[] | undefined> = {
+  road: ['cameras', 'ecameras'],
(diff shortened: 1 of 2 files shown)


incognitojam:
I think let's leave the aborting requests until tanstack query is in, it'll be a lot cleaner

I think just try resetting the component state for now

greatgitsby:
you're right, never used `AbortSignal` before. this will work in the meantime

greatgitsby:
this technically won't cancel the in-flight request, but it alleviates the UI inconsistency

src/components/RouteUploadButtons.tsx
```diff
@@ -63,7 +61,17 @@ const RouteUploadButtons: VoidComponent<RouteUploadButtonsProps> = (props) => {
     } as Record<ButtonType, 'idle' | 'loading' | 'success' | 'error'>,
   })
 
-  const updateButtonStates = (types: ButtonType[], state: 'loading' | 'success' | 'error') => {
+  createEffect(() => {
+    if (props.route?.fullname === currentRoute()?.fullname) return
```
incognitojam, line 65: You don't need a signal for this - effects will only run when their dependencies change

There's a helper function `on` you can use to explicitly define dependencies - https://docs.solidjs.com/reference/reactive-utilities/on-util

Keep in mind though that you are reading uploadStore.states so that probably adds a dependency - you could just create a list of the button types instead

incognitojam (changes requested):

incognitojam:
I just made some cleanups in #339 as well

PR #340, opened 2025-03-30
fix: reset upload btn state on new route selection

greatgitsby:
same purpose as https://github.com/commaai/connect/pull/338 (GH closed my PR when i rebased and dropped the commit)

https://github.com/user-attachments/assets/31471d8a-cc93-492e-b880-04a0feeb4c62


incognitojam:
I found a better way to write the file type logic: #351

greatgitsby:
that looks a lot better. thanks for doing that!

PR #341, opened 2025-03-30
feat(upload queue): optimistic cancel

greatgitsby:


sshane:
why is this so big?

incognitojam:
I think just not rebased after merging upload queue refactor

PR #352, opened 2025-03-30
fix: concat usage

greatgitsby:
introduced in https://github.com/commaai/connect/commit/0507b64e96f8009727ec871610e69353a1768676

buttons appear to not work but we aren't requesting any files types to upload when the helper is called

`concat()` does not modify underlying arrays


incognitojam:
thanks for spotting!

PR #356, opened 2025-03-30
feat: Group cards by date

ugtthis:
My attempt on Shane's idea PR #310. <br>
I like this idea since it makes cards easier to read and also makes card thinner taking up less screen real estate. I avoided duplicate headers by tracking which date headers we've already shown in a persistent Set and only displaying a header when it's the first time we encounter that specific date.

### DIFF
--- a/src/pages/dashboard/components/RouteList.tsx
+++ b/src/pages/dashboard/components/RouteList.tsx
@@ -37,11 +37,9 @@ const RouteCard: VoidComponent<RouteCardProps> = (props) => {
     <Card class="max-w-none" href={`/${props.route.dongle_id}/${props.route.fullname.slice(17)}`} activeClass="md:before:bg-primary">
       <CardHeader
         headline={
-          <div class="flex gap-2">
-            <span>
-              {startTime().format('h:mm A')} to {endTime().format('h:mm A')}
-            </span>
-          </div>
+          <span>
+            {startTime().format('h:mm A')} to {endTime().format('h:mm A')}
+          </span>
         }
         subhead={location()}
         trailing={
--- a/src/pages/dashboard/components/RouteList.tsx
+++ b/src/pages/dashboard/components/RouteList.tsx
@@ -9,8 +9,12 @@ import RouteStatistics from '~/components/RouteStatistics'
 import { getPlaceName } from '~/map/geocode'
 import type { RouteSegments } from '~/types'
 
+// Track rendered dates at module level to persist between renders
+const seenDates = new Set<string>()
+
 interface RouteCardProps {
   route: RouteSegments
+  showDateHeader: boolean
 }
 
 const RouteCard: VoidComponent<RouteCardProps> = (props) => {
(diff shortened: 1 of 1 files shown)


greatgitsby:
looks great!

src/pages/dashboard/components/RouteList.tsx
```diff
@@ -32,32 +35,34 @@ const RouteCard: VoidComponent<RouteCardProps> = (props) => {
...
-            <Show when={timeline()?.userFlags}>
-              <div class="flex items-center justify-center rounded-full bg-gradient-to-br from-amber-500 to-amber-900 p-2 border border-amber-300 shadow-inner shadow-black/20">
-                <Icon class="text-yellow-300" size="20" name="flag" filled />
-              </div>
-            </Show>
-          </Suspense>
-        }
-      />
+    <>
+      {props.shouldShowDateHeader && <h2 class="text-xl font-bold mt-6 pl-2">{startTime().format('ddd, MMM D, YYYY')}</h2>}
```
greatgitsby, line 39: can we move the margin elsewhere? the spacing looks odd on the first entry

sshane:
Doesn't work when you load a new page
Screenshot_20250330-145209

ugtthis:
Strange how Im not seeing the same things as you. I was loading new pages by scrolling down and haven't seen any duplicates yet. Been using my connect account since I have more routes, but also includes clips of the demo account below.

@greatgitsby you are seeing double headers also?


# using my connect account

### on Desktop

https://github.com/user-attachments/assets/4dae32f3-d0dc-41a7-8738-042c003ff1b2



---
# Using demo account

### on Desktop
https://github.com/user-attachments/assets/f56b4392-6376-4609-a6fb-9837212d2d10


### on iPhone
https://github.com/user-attachments/assets/98165278-08f3-4a85-acbf-ecd4b0d06495

greatgitsby:
try your account on iphone just in case.

i reproduced it on safari iOS:

image

ugtthis:
Thanks! For some reason it seems to be only a safari thing. I realized I was using chrome and firefox for both my macbook and iphone and didn't try safari.

sshane:
It's not browser specific, it just happens less often with devices with routes spaced out in time (like the demo device)

sshane:
I feel like the other cards shouldn't be aware of other cards, we should handle this all in RouteList

greatgitsby:
+1

think [groupBy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy) (or another similar pattern)

src/pages/dashboard/components/RouteList.tsx
ugtthis: I changed it in this commit https://github.com/commaai/connect/pull/356/commits/fa281e1204661d9288c2323be8c34ebb9e6bcd5f

ugtthis:
- removed top margin on first header
- adjusted date name to `'dddd, MMM D' : 'dddd, MMM D, YYYY'`
- parent manages header display

sshane:
This just seems to be a slightly different version of my solution.

src/pages/dashboard/components/RouteList.tsx
```diff
@@ -9,6 +9,8 @@ import RouteStatistics from '~/components/RouteStatistics'
 import { getPlaceName } from '~/map/geocode'
 import type { RouteSegments } from '~/types'
 
+const shownDateHeaders = new Set<string>()
```
sshane, line 12: Always minimize scope

PR #382, opened 2025-03-31
chore: generate PWA assets during build

incognitojam:
Generate manifest, images and splash screens, and inject necessary links into the html header


sshane:
Would prefer to not add any more dependencies until we ship. Is this absolutely needed? Old connect doesn't have

incognitojam:
old connect didn't keep up to date with the required splash screens (you have to specify one for every device width/height/pixel ratio that apple ships), and had all of the images checked in to the repo. this lets us generate all of the splash screens from the favicon svg

PR #396, opened 2025-04-01
fix: glitches when draggin cursor to end

greatgitsby:
resolves: https://github.com/commaai/connect/issues/390

https://github.com/user-attachments/assets/c5b94916-9012-4b7a-a89a-780faecbdae9


sshane:
Does this remove looping?

greatgitsby:
it does. the other fix requires telling the video about timeline touch events (if scrubbing, turn off looping until we stop holding mouse/finger down), which will require refactoring both components

PR #397, opened 2025-04-01
fix: video actually scrub to end

greatgitsby:
when paired with https://github.com/commaai/connect/pull/396, it's nice and predictable

https://github.com/user-attachments/assets/e44ff076-1d74-440a-aa59-f49f92fbfab5


greatgitsby:
don't merge until https://github.com/commaai/connect/pull/396 is merged

the bug is hidden w/o being able to go to end of video

PR #399, opened 2025-04-01
fix: snappier route load

greatgitsby:
partially resolves: https://github.com/commaai/connect/issues/262

### DIFF
--- a/src/pages/dashboard/activities/RouteActivity.tsx
+++ b/src/pages/dashboard/activities/RouteActivity.tsx
@@ -1,5 +1,6 @@
-import { createResource, createSignal, Match, Suspense, Switch, type VoidComponent } from 'solid-js'
+import { createEffect, createResource, createSignal, Match, Suspense, Switch, type VoidComponent } from 'solid-js'
 import { Navigate } from '@solidjs/router'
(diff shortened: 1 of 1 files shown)


greatgitsby:
need to fix tests. also may loop in the other requests to use improved query if it doesn't bloat this PR

src/App.browser.test.tsx
```diff
@@ -14,7 +15,16 @@ const SKIP_PRIVATE = true
 
 const UPLOAD_QUEUE = 'Upload Queue'
 
-const renderApp = (location: string) => render(() => <Routes />, { location, wrapper: AppLayout })
+const testWrapper = () => {
+  const queryClient = new QueryClient()
+  return (
+    <QueryClientProvider client={queryClient}>
+      <AppLayout />
```
incognitojam, line 22: need props.children inside AppLayout

sshane (commented):
Let me work through what's needed in https://github.com/commaai/connect/pull/392 first before we merge any individual PRs

sshane:
also tanstack might be a feature!

greatgitsby:
if it fixes the bug, you could argue it isn't, but i get the sentiment 😄

i think we should try fixing how we use resources first

PR #400, opened 2025-04-01
RouteCard suspense

incognitojam:
- re-write to not have chained resources
- location() should have a suspense around it (unless we put a suspense around each RouteCard instead)

### DIFF
--- a/src/pages/dashboard/components/RouteList.tsx
+++ b/src/pages/dashboard/components/RouteList.tsx
@@ -16,20 +16,17 @@ interface RouteCardProps {
 const RouteCard: VoidComponent<RouteCardProps> = (props) => {
   const startTime = () => dayjs(props.route.start_time_utc_millis)
   const endTime = () => dayjs(props.route.end_time_utc_millis)
-  const startPosition = () => [props.route.start_lng || 0, props.route.start_lat || 0] as number[]
-  const endPosition = () => [props.route.end_lng || 0, props.route.end_lat || 0] as number[]
-  const [startPlace] = createResource(startPosition, getPlaceName)
-  const [endPlace] = createResource(endPosition, getPlaceName)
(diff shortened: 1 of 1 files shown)


src/pages/dashboard/components/RouteList.tsx
```diff
@@ -42,7 +39,7 @@ const RouteCard: VoidComponent<RouteCardProps> = (props) => {
             </span>
           </div>
         }
-        subhead={location()}
+        subhead={<Suspense>{location()}</Suspense>}
```
greatgitsby, line 42: needs a fallback, might cause card drawing issues?

incognitojam: it just hides it when there's no fallback

greatgitsby: if subhead has a fixed height the. i'm not concerned

src/pages/dashboard/components/RouteList.tsx
```diff
@@ -16,20 +16,17 @@ interface RouteCardProps {
...
+  const [location] = createResource(async () => {
+    const startPos = [props.route.start_lng || 0, props.route.start_lat || 0]
+    const endPos = [props.route.end_lng || 0, props.route.end_lat || 0]
+    const startPlace = await getPlaceName(startPos)
+    const endPlace = await getPlaceName(endPos)
+    if (!startPlace && !endPlace) return ''
+    if (!endPlace || startPlace === endPlace) return startPlace
+    if (!startPlace) return endPlace
+    return `${startPlace} to ${endPlace}`
+  })
```
sshane, line 29: Won't everything be recomputed when one prop changes?

src/pages/dashboard/components/RouteList.tsx
sshane: I think we should add a fallback almost always

PR #402, opened 2025-04-01
refactor: create Sentinel component

incognitojam:


incognitojam:
seems to work the same

PR #403, opened 2025-04-01
RouteList: remove duplicate event.json queries

sshane:
Split from https://github.com/commaai/connect/pull/392

Reduces the span of time we're spending on json queries for a route of 12 segs from 35ms to 10ms (from first query queue to last query start)

### DIFF
--- a/src/components/RouteStatistics.tsx
+++ b/src/components/RouteStatistics.tsx
@@ -11,7 +11,7 @@ const formatEngagement = (timeline?: TimelineStatistics): string | undefined => 
   return `${(100 * (engagedDuration / duration)).toFixed(0)}%`
 }
 
-const RouteStatistics: VoidComponent<{ class?: string; route?: Route; timeline?: TimelineStatistics }> = (props) => {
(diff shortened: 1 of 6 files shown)


src/components/RouteStatistics.tsx
```diff
@@ -12,16 +11,14 @@ const formatEngagement = (timeline?: TimelineStatistics): string | undefined =>
   return `${(100 * (engagedDuration / duration)).toFixed(0)}%`
 }
 
-const RouteStatistics: VoidComponent<{ class?: string; route?: Route }> = (props) => {
-  const [timeline] = createResource(() => props.route, getTimelineStatistics)
-
+const RouteStatistics: VoidComponent<{ class?: string; route?: Route; timeline?: TimelineStatistics }> = (props) => {
```
incognitojam, line 14: We use this component in `<RouteActivity>` too, but because the prop is optional there is no error

incognitojam:
```suggestion
const RouteStatistics: VoidComponent<{ class?: string; route: Route | undefined; timeline: TimelineStatistics | undefined }> = (props) => {

incognitojam: I think that would cause an error to be raised?

PR #406, opened 2025-04-01
Remove non-optional optional keyword

sshane:


sshane:
@incognitojam know if there's any test to check for always-specified optionals so we get proper type checking?

PR #423, opened 2025-04-02
fix: reset seek time on route change

TheSecurityDev:
Fixes an issue (#425) where the video seek and timeline don't reset back to the beginning when you change routes while playing. It still supports loading the start time from the URL path.


sshane:
This is adding more lines to fix a regression, we should just revert

PR #428, opened 2025-04-02
fix: handle video stream unavailable

greatgitsby:
resolves part of: https://github.com/commaai/connect/issues/264

handle the state where qcam isn't uploaded yet or a fatal network error occurs (testing local shows it tries to load ~3 times before aborting)

if the video fails to load, we disable the timeline interactivity and hide the cursor to avoid the impression it can be interacted with.

msg is copied from old connect

https://github.com/user-attachments/assets/e6b6c334-72d5-4264-8653-0e5e069f8a2d

## iOS

image


greatgitsby:
why does bundle size increase over 90kb??

greatgitsby:
30% of the added lines are from formatting.

src/components/Timeline.tsx
```diff
@@ -150,35 +151,39 @@ const Timeline: VoidComponent<TimelineProps> = (props) => {
...
-  createEffect(() => {
-    setMarkerOffsetPct((props.seekTime / duration()) * 100)
-  })
+  createEffect(() => setMarkerOffsetPct((props.seekTime / duration()) * 100))
 
   return (
     <div class="flex flex-col">
       <div class="h-1 bg-surface-container-high">
-        <div class="h-full bg-white" style={{ width: `calc(${markerOffsetPct()}% + 1px)` }} />
+        <div class={clsx('h-full bg-white', !props.interactive && 'hidden')} style={{ width: `calc(${markerOffsetPct()}% + 1px)` }} />
```
sshane, line 172: can we remove all this?

src/pages/dashboard/activities/RouteActivity.tsx
```diff
@@ -57,8 +61,22 @@ const RouteActivity: VoidComponent<RouteActivityProps> = (props) => {
...
+            startTime={seekTime()}
+          />
+          <Timeline
+            class="mb-1"
+            interactive={timelineIsInteractive()}
+            route={route.latest}
+            seekTime={seekTime()}
+            updateTime={onTimelineChange}
+            events={events()}
+          />
```
sshane, line 79: same

greatgitsby: adding the new props makes Biome wrap everything. i can override the style...

greatgitsby: rename to `interactive` helped here. RoutePlayer i simplified

src/components/Timeline.tsx
greatgitsby: sure! it actually looks more consistent if the marker is always visible

src/components/Timeline.tsx
```diff
@@ -170,7 +174,8 @@ const Timeline: VoidComponent<TimelineProps> = (props) => {
       <div
         ref={ref!}
         class={clsx(
-          'relative isolate flex h-8 cursor-pointer touch-none self-stretch rounded-b-md bg-blue-900',
+          'relative isolate flex h-8 touch-none self-stretch rounded-b-md bg-blue-900',
+          props.interactive && 'cursor-pointer',
```
greatgitsby, line 178: keeping this so on desktop it doesn't make users think its clickable

src/components/RouteVideoPlayer.tsx
```diff
@@ -10,6 +10,7 @@ type RouteVideoPlayerProps = {
   class?: string
   routeName: string
   startTime: number
+  onEvent?: (event: 'load' | 'error') => void
```
incognitojam, line 13: I think any props we expect to use in other parts of the app we shouldn't make optional

Just have optional props on reusable UI components like buttons, cards etc

sshane:
We should handle routes with missing qcams, so removing the timeline lockout when we see an error

sshane:
- [ ] doesn't reset time when clicking off error'd route
- [ ] duration is previous route duration (reset to events or 0:00)

src/pages/dashboard/activities/RouteActivity.tsx
```diff
@@ -42,6 +42,12 @@ const RouteActivity: VoidComponent<RouteActivityProps> = (props) => {
     if (video) video.currentTime = newTime
   }
 
+  createEffect(() => {
+    routeName() // track changes
+    setSeekTime(props.startTime)
+    onTimelineChange(props.startTime)
+  })
+
```
sshane, line 50: split this out

src/components/RouteVideoPlayer.tsx
```diff
@@ -98,21 +103,32 @@ const RouteVideoPlayer: VoidComponent<RouteVideoPlayerProps> = (props) => {
...
+          if (!data.fatal) return
+          if (data.type === ErrorTypes.NETWORK_ERROR) onError()
+        })
         player.attachMedia(video)
         setHls(player)
       })
       onCleanup(() => hls()?.destroy())
     } else {
       setHls(null)
+      video.onerror = onError
```
sshane, line 120: @incognitojam I tried to use the non-hls player to test but it never loads

src/pages/dashboard/activities/RouteActivity.tsx
greatgitsby: this is to fix if vid can't load, marker is not at zero, and you switch to a new video

it resets the marker to t=0

greatgitsby: necessary for this PR

src/components/RouteVideoPlayer.tsx
greatgitsby: image

are you referring to this?

safari on iOS doesn't use hls player. if i don't have this line, it doesn't error out

src/pages/dashboard/activities/RouteActivity.tsx
sshane: I mean merge it first. It's a separate fix for slow internet as well

greatgitsby: i'm tracking. will do

greatgitsby: https://github.com/commaai/connect/pull/478

sshane:
I just checked and master doesn't currently handle routes with half missing qcams, so that's not a requirement for this PR. We need it soon though!

src/components/RouteVideoPlayer.tsx
sshane: changing to event listener to be consistent

sshane: just saying I couldn't get it to play on Chrome, but good that it works on iOS

greatgitsby:
the reset of error needs to be more optimistic than on loaded video:

https://github.com/user-attachments/assets/4e167cb7-9000-482a-b4d3-b60b8ff20cc2

sshane:
done

https://github.com/user-attachments/assets/51ce341e-1c02-4497-9314-6d192888bddb

PR #429, opened 2025-04-02
refactor: Timeline in RouteVideoPlayer

greatgitsby:
RE: https://github.com/commaai/connect/pull/428

these components are tightly coupled, why not combine them? adding other bug fixes will require these components to talk even more to each other, it'll be simpler in one scope.

### DIFF
--- a/src/components/RouteVideoPlayer.tsx
+++ b/src/components/RouteVideoPlayer.tsx
@@ -11,10 +11,10 @@ import { TimelineEvent } from '~/api/derived'
 
 type RouteVideoPlayerProps = {
   class?: string
+  events: TimelineEvent[]
+  route: Route | undefined
   routeName: string
   startTime: number
-  route: Route | undefined
-  events: TimelineEvent[]
 }
 
 const RouteVideoPlayer: VoidComponent<RouteVideoPlayerProps> = (props) => {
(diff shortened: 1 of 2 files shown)


src/components/RouteVideoPlayer.tsx
```diff
@@ -5,16 +5,20 @@ import { getQCameraStreamUrl } from '~/api/route'
...
+import { Route } from '~/api/types'
+import { TimelineEvent } from '~/api/derived'
 
 type RouteVideoPlayerProps = {
   class?: string
   routeName: string
   startTime: number
-  onProgress: (seekTime: number) => void
-  ref: (el?: HTMLVideoElement) => void
+  route: Route | undefined
```
greatgitsby, line 16: i did this because @sshane started using this in prop types

my guess as to why is that `route?: Route` denotes *optional* as in "you can leave it out **if you want**" vs. `route: Route | undefined` denotes *optional* in the sense that it might be missing **due to an error**.

is that the line of thinking?

sshane: I just learned this too. Yes, `route?: Route` is like `route=None` in Python

greatgitsby: thanks!

sshane: btw if it's a reactive object like a resource it can also be undefined on first component load if I understand correctly, then any createResource or createEffect, etc. inside the child component are called when the reactive property loads.

https://github.com/commaai/connect/blob/8c99b76899b77b9a0c988f0b08e2a2a3e55653a9/src/components/RouteStaticMap.tsx#L55-L57

sshane (commented):
We may need the timeline in RouteActivity to make the segment selection work and updating the copy button, etc, cc @incognitojam

greatgitsby:
fair point... i might have to just make do passing callbacks around.

greatgitsby:
i'll make do without

PR #435, opened 2025-04-02
Remove duplicate devices query when switching device

sshane:
I would like to merge this instead of https://github.com/commaai/connect/pull/430, but we need to decide how to handle public routes.

### DIFF
--- a/src/api/devices.ts
+++ b/src/api/devices.ts
@@ -14,38 +14,8 @@ const sortDevices = (devices: Device[]) =>
     }
   })
 
-// export const SHARED_DEVICE = 'Shared Device'
-
-// const createSharedDevice = (dongleId: string): Device => ({
-//   dongle_id: dongleId,
-//   alias: SHARED_DEVICE,
-//   serial: '',
-//   last_athena_ping: 0,
-//   ignore_uploads: null,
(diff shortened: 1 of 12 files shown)


src/api/devices.ts
```diff
@@ -14,38 +14,8 @@ const sortDevices = (devices: Device[]) =>
     }
   })
 
-export const SHARED_DEVICE = 'Shared Device'
-
-const createSharedDevice = (dongleId: string): Device => ({
```
sshane, line 19: We can come back and re-add  if we decide we really want to show in the sidebar when viewing a shared device, but we're already about to remove the route listing, so probably not a big deal to match old connect

sshane:
@incognitojam do you know why this expects `demo 3X` to not be there when in demo mode?

https://github.com/commaai/connect/blob/e2d2d5978a10bcbcf64efc652337d5a49ab6bacd/src/App.browser.test.tsx#L25

incognitojam:
It says not false!
It was like that in the vitest template and I just never changed it

sshane:
Now it loads the device name immediately for DeviceActivity, and the current check asserts only 1 demo 3X string

PR #436, opened 2025-04-02
Remove shared device object

sshane:
Split from https://github.com/commaai/connect/pull/435, just a clean up

### DIFF
--- a/src/api/devices.ts
+++ b/src/api/devices.ts
@@ -14,41 +14,7 @@ const sortDevices = (devices: Device[]) =>
     }
   })
 
-// export const SHARED_DEVICE = 'Shared Device'
-
-// const createSharedDevice = (dongleId: string): Device => ({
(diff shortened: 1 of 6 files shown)


src/api/devices.ts
```diff
@@ -14,39 +14,7 @@ const sortDevices = (devices: Device[]) =>
     }
   })
 
-export const SHARED_DEVICE = 'Shared Device'
```
sshane, line 17: We can come back and re-add if we decide we really want to show in the sidebar when viewing a shared device, but we're already about to remove the route listing, so probably not a big deal to match old connect

incognitojam: ACK

sshane:
This broke https://github.com/commaai/connect/pull/430 (from the memo?), but https://github.com/commaai/connect/pull/435 properly fixes it

PR #439, opened 2025-04-02
fix: no redraw on device switch

greatgitsby:


greatgitsby:
way more buggy. child components need stabilizing before this is a good idea

PR #440, opened 2025-04-02
cleanup: combined url state

incognitojam:


src/pages/dashboard/Dashboard.tsx
```diff
@@ -104,22 +104,23 @@ const DashboardLayout: Component<{
...
 const Dashboard: Component<RouteSectionProps> = () => {
   const location = useLocation()
-
-  const pathParts = () => location.pathname.split('/').slice(1).filter(Boolean)
-  const dongleId = () => pathParts()[0]
-  const dateStr = () => pathParts()[1]
-  const startTime = () => (pathParts()[2] ? Number(pathParts()[2]) : 0)
-
-  const pairToken = () => !!location.query.pair
+  const urlState = () => {
```
greatgitsby, line 107: can memoize

sshane:
Let's make it redirect back to the default dongle if you navigate to a dongle ID that doesn't exist in your devices, or a route ID that doesn't return anything

PR #442, opened 2025-04-02
feat: cache derived results

incognitojam:
just a memory cache, should de-duplicate requests

### DIFF
--- a/src/api/derived.ts
+++ b/src/api/derived.ts
@@ -90,8 +90,21 @@ const getDerived = async <T>(route: Route, fn: string): Promise<T[]> => {
   return (await Promise.all(results)).filter((it) => it !== undefined)
 }
 
-export const getCoords = (route: Route): Promise<GPSPathPoint[]> =>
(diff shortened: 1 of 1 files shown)


greatgitsby:
https://github.com/commaai/connect/pull/448 might handle this without any additional code, what do you think?

greatgitsby:
the manifest one is simply switching to createQuery, it's 3 lines

sshane:
As I said somewhere else, I think there's two ways to de-duplicate requests, but we should follow some priority:

1. share state first, i.e. pass in properties, global signals(?), etc. within reason.
2. if for some reason that's too complex, then we can cache the queries

If we do 2 for one query, we need to make sure we aren't querying it twice in two different places needlessly, or caching may not help (like in RouteList and in RouteActivity if you visit a route URL)

incognitojam:
I'm not sure I understand the last point, in this PR sharing the promise means there won't be two requests concurrently

sshane:
Ah, if it's all single-threaded, it should be safe then

sshane:
Let's only work on stuff that improves the worst case experience for now, like all the duplicate queries!

PR #443, opened 2025-04-02
fix: Route chained resources [wip]

incognitojam:
still not instant


sshane:
done in https://github.com/commaai/connect/pull/452

PR #446, opened 2025-04-03
fix: remove this top level suspense

greatgitsby:
this gets rid of a ton of weird issues for me


sshane:
Nice find, child Suspense's seem to work much more reliably.

Need to fix some bad and missing fallbacks now (like Shared Device)

/home/batman/Videos/clips/less-suspense.mp4

greatgitsby:
`solid-query` can create [really simple dependent queries](https://github.com/commaai/connect/pull/447/files#diff-b2bd04aca6efa28fe7a42ac83d8d75e7527ddacf3e9479742c850a30494c0fadR41)

only needed if they truly are dependent

greatgitsby:
opening as this solves some of our suspense bugs

PR #448, opened 2025-04-03
refactor: fast routes v2 [wip]

greatgitsby:
needs more error handling but it no longer blocks


sshane:
Very quick! But the RouteActivity's elements jump around upon switching, and it looks like it's using some sort of caching per-page-load.

As usual the fix is a red diff!

https://github.com/commaai/connect/pull/452 and your https://github.com/commaai/connect/pull/446

greatgitsby:
not sure what you mean about jumping? the skeleton loader? the sizing is preserved in the suspended state

PR #451, opened 2025-04-03
Add loading animation to video player

sshane:


sshane:
On mobile I was just able to repro it not loading videos with an infinite animation, but then I tried again and can't repro, so not sure if this PR regresses anything...

sshane:
Could've been a route with no qcams, need an error message!

PR #452, opened 2025-04-03
Instant route switching

sshane:
Simple version of https://github.com/commaai/connect/pull/451 without any thought on fancy animations/quirky transitions

https://github.com/user-attachments/assets/d3c5731f-c539-43cf-94e6-aecc78aba1b0

Enabled by https://github.com/commaai/connect/pull/446

### DIFF
--- a/src/components/RouteStatistics.tsx
+++ b/src/components/RouteStatistics.tsx
@@ -6,7 +6,7 @@ import { formatDistance, formatRouteDuration } from '~/utils/format'
 import StatisticBar from './StatisticBar'
 
 const formatEngagement = (timeline: TimelineStatistics | undefined): string | undefined => {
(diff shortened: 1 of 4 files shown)


greatgitsby:
simple! the switch is instant in RouteList, but it still feels like it hangs, more evident when you add network latency

the suspense actually helps make the UI feel quicker to use

https://github.com/user-attachments/assets/ddd661b6-c807-41c6-95f4-b55b1c653ba0

sshane:
Yes, the only thing we know is the route ID, so that changes instantly. The rest we can add animations for to make it look a little better

sshane:
There was some flashing on all the RouteList RouteStatistics, not sure if that was always a bug or not, hopefully we can just get and compute the timeline events once soon

PR #455, opened 2025-04-03
Fix flashing when clicking around

sshane:
Closes https://github.com/commaai/connect/issues/454

Not sure why the RouteStatistics animation in the RouteActivity is gone again, but oh well

@incognitojam we can probably remove a bunch of Suspense's. Maybe the rule of thumb is; you don't want a Suspense without a fallback?


greatgitsby:
the entire reason for using Suspense is to have a fallback element that gets swapped out once the data is ready

using it without is undefined behavior

PR #18, opened 2024-06-12
Fixed overflow when scrolling up

ugtthis:
### Sidebar showing if scrolled up too much
ui-overflow

This shows up for both mobile and desktop.

### First one shows the overflow second shows it fixed

https://github.com/commaai/new-connect/assets/142481257/801b3ad6-ea1a-444c-8a8c-52c6bbde3c45

PR #459, opened 2025-04-03
this?

sshane:


greatgitsby:
it works in the browser but not the PWA. seems [intentional](https://stackoverflow.com/a/76475382)?

there should be a better way than installing a random package

incognitojam:
Looks like we had a custom component for this in old connect - https://github.com/commaai/connect-pwa-archive/pull/249

I think they disable it on iOS PWA when you pick "standalone" to make it feel more like a native app

incognitojam:
Trying this PR on iOS/Safari, if I drag the page down it does overscroll, but it doesn't trigger a refresh. It also breaks scrolling the page normally..

incognitojam:
I think I narrowed down the fix for Android to just one change

incognitojam:
Hm we added that class to fix another bug on iOS https://github.com/commaai/connect/pull/18

sshane:
This is huge. First let's do the simplest thing on Android without regressing, then open another PR for iOS.

This is not splitting PRs up

src/index.css
```diff
@@ -120,7 +120,10 @@
      * fill visible viewport
      * https://developers.google.com/web/updates/2016/12/url-bar-resizing
      */
-    @apply h-full overflow-hidden;
+    @apply h-full;
+
+    /* prevent horizontal scrolling, e.g. when drawer is open */
+    @apply overflow-x-hidden;
```
sshane, line 126: I don't notice this, what is it for?

src/components/material/Drawer.tsx
```diff
@@ -48,7 +48,8 @@ const Drawer: ParentComponent<DrawerProps> = (props) => {
         class="hide-scrollbar fixed inset-y-0 left-0 h-full touch-pan-y overflow-y-auto overscroll-y-contain transition-drawer duration-500"
         style={{
           left: drawerVisible() ? 0 : `${-PEEK}px`,
-          opacity: drawerVisible() ? 1 : 0.5,
+          // Opacity should be 0 when drawer is closed, otherwise it can be visible when overscrolling
+          opacity: drawerVisible() ? 1 : 0,
```
sshane, line 52: Is this for iOS? If so, what was all the other stuff for?

src/index.css
incognitojam: You'd have to test the branch as it was in be22dc8. We have to clip the overflow horizontally because we have a couple of animations which translate the UI horizontally:
- Opening the drawer slides it from left to right slightly, and the rest of the UI slides to the right
- The route activity slides away to the right when you go back to the route list
Without this you can scroll the whole page horizontally.

src/components/material/Drawer.tsx
incognitojam, line 52: This is for the bug previously addressed in #18

incognitojam:
Even this change causes weirdness on iOS: when I open the drawer, if I swipe left I can scroll and see the dashboard again. I can't repro it on the deployed master.

sshane:
Can we merge something simple that works on Android?

PR #464, opened 2025-04-03
fix: ios pwa splash screen

incognitojam:


incognitojam:
So adding this meta tag to the header fixes splash screens on iOS:
```html
<meta name="apple-mobile-web-app-capable" content="yes">
```

But web.dev says not to use it as it's deprecated
https://web.dev/learn/pwa/web-app-manifest/#designing-your-pwa-experience
image

PR #467, opened 2025-04-03
fix: remove extraneous SuspenseList in Dashboard

incognitojam:
accidentally leftover in #160

SuspenseList controls the order in which content is revealed (when the fallback is removed), so it would appear that some Suspense were triggered when they were just blocked by another Suspense (depending on direction of `revealOrder`)


sshane:
Can you quickly click around and make sure nothing is broken and note any changes (with Fast 3G)? LGTM

incognitojam:
I think it's good

incognitojam:
nvm did a proper side by side:
- you can see the `-` fallbacks in the route cards

sshane:
fine to merge and we can fix in a follow up, it's not expected to be perfect right now

sshane:
> not sure why the duration and distance don't load instantly still since they come from the route()

we do another getRoute query needlessly on entrance

sshane:
@incognitojam this is exactly what I was mentioning, please test before merging. can you fix?

image

PR #470, opened 2025-04-04
fix: show route distance and duration immediately

incognitojam:
This looks like it works fine now?


sshane:
already have https://github.com/commaai/connect/pull/468

PR #473, opened 2025-04-04
bg-surface-container-low is banned

sshane:


sshane:
idk, things look slightly uglier. should add gradients and copy old connect's color scheme first

PR #474, opened 2025-04-04
fix: RouteActivity suspense not blocking any longer

greatgitsby:
suspense works in RouteActivity now. `keyed` makes Solid re-render the component from scratch, so in this case the `RouteActivity` is re-rendered from scratch. i think the re-use between navigations was causing issues with resources in `RouteActivity`

we need `keyed` at the `dongleId` above for a similar reason.

i think it's technically an escape hatch, but we can spend time later on removing this.

need to rework RouteStatistics, engaged shows NaN% on immediate load

https://github.com/user-attachments/assets/bb0670c6-a869-4c22-a5c6-0be73582c4ac

### DIFF
--- a/src/pages/dashboard/Dashboard.tsx
+++ b/src/pages/dashboard/Dashboard.tsx
@@ -145,8 +145,8 @@ const Dashboard: Component<RouteSectionProps> = () => {
                   <Match when={urlState().dateStr === 'settings' || urlState().dateStr === 'prime'}>
                     <SettingsActivity dongleId={dongleId} />
                   </Match>
-                  <Match when={urlState().dateStr}>
-                    {(dateStr) => <RouteActivity dongleId={dongleId} dateStr={dateStr()} startTime={urlState().startTime} />}
+                  <Match when={urlState().dateStr} keyed>
+                    {(dateStr) => <RouteActivity dongleId={dongleId} dateStr={dateStr} startTime={urlState().startTime} />}
                   </Match>
                 </Switch>
               }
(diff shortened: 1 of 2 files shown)


greatgitsby:
engaged is a bit broken but can be fixed in following pr

sshane:
Nice fix! Just have a couple thoughts

I liked it maintaining the previous route state before switching in some cases. For example, the timeline flashing from the previous route to a blank state (fallback) animation then immediately to the new timeline looks janky with fast internet, but better with slow internet. But this can be fixed by sharing the timeline events we already queried in the list.

Anything else you can think of that will look worse after this? Toggles?

sshane:
just to show what I mean. good with slow, iffy with fast:

https://github.com/user-attachments/assets/62d2bb6d-3444-415e-a2d9-be16069a3862

sshane:
I wonder if we can do something similar to replace the fast device switching stuff I did?

greatgitsby:
some thoughts:

the fundamental problem we have with having to do multiple calls in separate parts of the app is because of our hierarchy:
```
Dashboard
  -> DeviceActivity
    -> device() query
    -> RouteList
      -> page() query (N route()), N events() requests
  -> RouteActivity
    -> device()
    -> profile()
    -> route() query
    -> events() query (in getTimelineEvents)
    -> setRouteViewed()
    -> etc...
```

the events() query buried in the RouteList can't propagate data to RouteActivity in any clean way. we would have to use a wrapping Context component and manage it ourselves. or, we use an abstraction like tanstack which uses the same concept (queries are stored in the QueryContext which is basically an in memory cache).

we can do either. we would check the cache to see if the data has been fetched, is fresh, etc..

both options fix the underlying issue

my thought:
- tanstack uses queryKeys to store its data
- when we query the route list, we can store the route() we've encountered, same with events, device, etc.
- then, you click on a route, every query has already been seen, the RouteActivity loads immediately
- toggling between routes is instant w/o redraws because the data is already in the cache when we hit the query resource
- we maintain reactivity and don't have to use `keyed` anywhere

sshane:
What about a global signal that the route cards could set? It would be listened to in route activity.

greatgitsby:
🤔 that might work!

PR #482, opened 2025-04-04
Add Suspense animation back to Timeline

sshane:


sshane:
@greatgitsby wanna see how easy it will be to remove the route query and extra events.json queries?

greatgitsby:
see my comment here, i talk about that issue: https://github.com/commaai/connect/pull/474#issuecomment-2777535387

PR #483, opened 2025-04-04
refactor: zero duplicate queries

greatgitsby:
just a wip; store current metadata in a Context

won't end up merging this, will break it up if we like it

### DIFF
--- a/src/AppContext.tsx
+++ b/src/AppContext.tsx
@@ -1,0 +1,51 @@
+import { createContext, JSX, useContext } from 'solid-js'
+import { createStore } from 'solid-js/store'
+import { Device, Route } from './api/types'
+import { TimelineEvent, TimelineStatistics } from './api/derived'
+
+interface AppState {
+  currentDevice: Device | undefined
+  currentRoute: Route | undefined
+  currentEvents: TimelineEvent[] | undefined
+  currentTimelineStatistics: TimelineStatistics | undefined
+}
+
+const INITIAL_STATE: AppState = {
+  currentDevice: undefined,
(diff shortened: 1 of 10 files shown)


sshane:
Doesn't seem to load yet

greatgitsby:
its very broken rn, give me a few

greatgitsby:
loads now but it's way more buggy

the edge case is when the user loads a route directly. hacking the fallback to state is not working consistently. no to mention it's disgusting

greatgitsby:
i'll play with this idea some more later, or anyone's welcome to pull this and play with it locally

greatgitsby:
man, it absolutely rips though... this is the sort of speed we can expect with no dupe queries

src/AppContext.tsx
```diff
@@ -0,0 +1,55 @@
...
+  ] as const
+
+  return <AppContext.Provider value={app}>{props.children}</AppContext.Provider>
+}
+
+export const useAppContext = () => {
+  const context = useContext(AppContext)
+  if (!context) throw new Error('Could not find an AppContext!')
+  return context
+}
```
sshane, line 55: What does app contexts get us over global signals like https://github.com/commaai/connect/pull/503?

greatgitsby: flexible usage. you import this wherever you need it without having to pass signals down the component tree

greatgitsby: i left more analysis on your PR

sshane: We can do that with global signals AFAIK too

PR #493, opened 2025-04-04
feat: color route card based on time

greatgitsby:
my take on the route color, but attempting to be simpler, no deps (thx gpt)

pending:
- [x] test cases

image

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -80,7 +80,7 @@ export const formatDate = (input: dayjs.ConfigType): string => {
 
 export const dateTimeToColorBetween = (startTime: Date, startColor: number[], endColor: number[]): string => {
   const sunrise = 5 // hours
-  const sunset = 6 + 12
+  const sunset = 5 + 12
   const fade = 2
 
   const hours = startTime.getHours() + startTime.getMinutes() / 60
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -78,31 +78,17 @@ export const formatDate = (input: dayjs.ConfigType): string => {
   return date.format('MMMM Do' + yearStr)
 }
 
-const hexToRgb = (hex: string): [number, number, number] => {
-  hex = hex.replace('#', '')
-  if (hex.length !== 6) throw new Error('Invalid hex color')
-  const [r, g, b] = [parseInt(hex.slice(0, 2), 16), parseInt(hex.slice(2, 4), 16), parseInt(hex.slice(4, 6), 16)]
-  return [r, g, b]
-}
+export const dateTimeToColorBetween = (date: Date, startColor: string, endColor: string): string => {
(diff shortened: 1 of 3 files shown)


sshane:
Maybe we either dolor the time background or the timeline background if we add it?

greatgitsby:
eh time background looks worse, and has more edge cases on text color

image

greatgitsby:
i like timeline background if it gets added

sshane:
your first solution is better here

sshane:
Tried it out, this is pretty nice!

Can we make 4am routes look the same as 10pm routes? The light level is the same pretty much

image

sshane:
Can we also just do a simple linear interp from day to night?

sshane:
Can we do something like np.interp(time, [6am, 12pm, 6pm], [blue, orange, blue])? Note just the clipping is different than your solution

greatgitsby:
🤔 sure, i'll do some research

sshane:
Trying to get this to match our routes, but sunset/sunrise times can vary wildly. Right now it's bright at 7pm, but in Jan it gets dark at 5, for now we can assume some average.

What do we think about something like this?

image

greatgitsby:
i see you've went a different direction, but here's a modified version of something like what i had before but with clamping:

```
const toRGB = (hex: string): number[] => hex.match(/\w\w/g)!.map((x) => parseInt(x, 16))
const toHex = (rgb: number[]): string => rgb.map((x) => Math.round(x).toString(16).padStart(2, '0')).join('')
const midnight = toRGB('#191970')
const sunlight = toRGB('#fdfbd3')
const hours = date.getHours() + date.getMinutes() / 60;
const clipped = Math.min(Math.max(hours, 6), 18) // clamp to 6am-6pm
const progress = (clipped - 6) / 12 // normalize to 0-1
const factor = Math.sin(Math.PI * progress)
const blended = midnight.map((x, i) => Math.round(x + (sunlight[i] - x) * factor))
return `#${toHex(blended)}`
```

sshane:
I just tweaked it a bit to remove most of the in-between colors. With the thumbnails you hardly ever saw the 30 minute twilight transition. It's mostly ever day or night

image

src/utils/format.ts
```diff
@@ -77,3 +77,21 @@ export const formatDate = (input: dayjs.ConfigType): string => {
   const yearStr = date.year() === dayjs().year() ? '' : ', YYYY'
   return date.format('MMMM Do' + yearStr)
 }
+
+export const dateTimeToColorBetween = (startTime: Date, startColor: number[], endColor: number[]): string => {
+  const sunrise = 5 // hours
+  const sunset = 5 + 12
+  const fade = 2 // wide transition since this accounts for varying daylight
```
sshane, line 84:
```suggestion
  const fade = 2 // wide transition since this accounts for different seasons
```

PR #495, opened 2025-04-04
Add all time label to device stats

sshane:


src/pages/dashboard/activities/DeviceActivity.tsx
```diff
@@ -145,6 +145,9 @@ const DeviceActivity: VoidComponent<DeviceActivityProps> = (props) => {
             </div>
           </div>
           <Show when={isDeviceUser()}>
+            <div class="flex items-center justify-between p-4 pb-0">
+              <h3 class="text-label-md uppercase">All Time</h3>
```
greatgitsby, line 149: i think the convention is to use `span` and let tailwind handle text sizing, since header defines a certain size

as i grep the codebase, it looks like it's not super consistent though

PR #503, opened 2025-04-05
Instant route data loading

sshane:


src/pages/dashboard/activities/RouteActivity.tsx
```diff
@@ -85,24 +83,24 @@ const RouteActivity: VoidComponent<RouteActivityProps> = (props) => {
...
+            <RouteUploadButtons route={currentRoute()} />
           </div>
         </div>
 
         <div class="flex flex-col gap-2">
           <span class="text-label-md uppercase">Route Map</span>
           <div class="aspect-square overflow-hidden rounded-lg">
             <Suspense fallback={<div class="h-full w-full skeleton-loader bg-surface-container" />}>
-              <RouteStaticMap route={route()} />
+              <RouteStaticMap route={currentRoute()} />
```
sshane, line 103: @greatgitsby have any thoughts about this? Passing in global signal from higher up, or just let each child component import the signal as needed? Not sure what's the more common pattern in web dev

src/store.tsx
```diff
@@ -0,0 +1,7 @@
+import { createSignal } from 'solid-js'
+import type { RouteSegments, Device } from '~/api/types'
+
+// TODO: remove undefined
+export const [currentRoute, setCurrentRoute] = createSignal<RouteSegments | undefined>(undefined)
+export const [currentEvents, setCurrentEvents] = createSignal([]) // todo annotate this and change to base events
+export const [currentDevice, setCurrentDevice] = createSignal<Device | undefined>(undefined)
```
sshane, line 7: Might need to make resources for each if we need Suspense animations to work around these, but the only time I think we'll need that is when you visit a URL of a route and we need to load the route info from the moment you see the page.

src/pages/dashboard/activities/RouteActivity.tsx
greatgitsby: you're doing prop drilling which is one solution. the problem comes when you want to add more signals, you have to add more props, drill down further...

a Context abstracts all this away and you can reference whatever data you want whenever you need it. ends up being simpler to just call useAppContext() to get whatever you need

sshane: This was to get it working, I was thinking about just doing `<RouteUploadButtons />`, `<RouteStaticMap />` and import the getter in each component as needed

greatgitsby: oh interesting, i missed that you're creating global signal variables

i've always thought that global state is an antipattern in these types of programs, but if it solves the problem in less lines, it can always be refactored later

greatgitsby: i also don't know how signals behave in a global context... we'd need to test this out

src/store.tsx
greatgitsby: one idea: ditch Suspense everywhere. use Show when truthy else skeleton loader. use .latest to not read under suspense boundary

src/pages/dashboard/activities/RouteActivity.tsx
sshane: https://www.solidjs.com/tutorial/stores_nocontext

> https://www.solidjs.com/tutorial/stores_nocontext

src/store.tsx
sshane: suspense is banned!

sshane:
We should be able to make a signal with all the loaded routes, then check once url state date string changes and set current route from that

greatgitsby:
check my draft PR; set an event on the route card click to setCurrentRoute, route is a prop

sshane:
This is for when you visit a URL to a route and don't click on the card

src/store.tsx
incognitojam: what?

greatgitsby: it's not that it's confusing. trying to use a signal that holds a Route and also using a resource for fetching a Route doesn't play well together -- that's why i proposed removing it as *a* solution

incognitojam:
Partially implemented this in #579, but haven't tackled fetching the route or events

PR #512, opened 2025-04-06
fix: redirect user if they arent logged in

greatgitsby:
fixes point 1 of https://github.com/commaai/connect/issues/506

we could debate the logged out experience (maybe user should be able to view public route even if they aren't logged in), but that's a design conversation. this solves the current issue

https://github.com/user-attachments/assets/2abbe7e5-6c09-4d1c-888e-2804c6d8fee1

### DIFF
--- a/src/pages/dashboard/Dashboard.tsx
+++ b/src/pages/dashboard/Dashboard.tsx
@@ -3,7 +3,7 @@ import type { Component, JSXElement, VoidComponent } from 'solid-js'
 import { Navigate, type RouteSectionProps, useLocation } from '@solidjs/router'
 import clsx from 'clsx'
 
-import { getAccessToken } from '~/api/auth/client'
+import { isSignedIn } from '~/api/auth/client'
 import { USERADMIN_URL } from '~/api/config'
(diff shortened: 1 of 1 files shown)


src/pages/dashboard/Dashboard.tsx
```diff
@@ -126,6 +128,10 @@ const Dashboard: Component<RouteSectionProps> = () => {
     return devices()?.[0]?.dongle_id
   }
 
+  onMount(() => {
+    if (!getAccessToken()) navigate('/login')
```
incognitojam, line 132: We already have a `<Navigate href="/login">` case in the JSX switch below, it's just *after* we check whether we are visiting a specific route, so it only redirects visits to `/` to `/login`. We can remove that if we have this

I'm not sure there is a need to put this inside `onMount`, we should have the access token immediately

incognitojam:
Oh, I also think we have logic in the DashboardDrawer for the user profile card to handle them not being signed in which we could remove

sshane:
I believe this regressed when you get a 503 you are now force logged out

PR #514, opened 2025-04-06
refactor: retrieve routes using `solid-query`

greatgitsby:
here's another simple way to dedupe all route retrievals. when we've fetched a page, store the route in the tanstack query cache, and fetch the route in the `RouteActivity` using tanstack. this solves two major things:

1. Suspense behaves exactly as before (`routes.data` will only trigger suspense if the data wasn't previously recorded). see demo video and test yourself
2. handles navigating to a route directly smoothly. the data isn't in the cache yet (maybe RouteList isn't populated), but it still fetches w/o any special handling, and Suspense is triggered for route metadata like stats in `RouteActivity`

i think introduction of `solid-query` at this point is going to help solve the remaining performance requirements we have, like deduping requests and *eventually* introducing metadata caching like old connect. but before that, no unnecessary requests!

looking for feedback and opinions on the approach

https://github.com/user-attachments/assets/0c59ca45-b499-4115-99e6-349a58778f1c

### DIFF
--- a/src/App.browser.test.tsx
+++ b/src/App.browser.test.tsx
@@ -4,10 +4,30 @@ import { configure, render, waitFor } from '@solidjs/testing-library'
 import { clearAccessToken, setAccessToken } from '~/api/auth/client'
 import * as Demo from '~/api/auth/demo'
 import { AppLayout, Routes } from './App'
+import { QueryClient, QueryClientProvider } from '@tanstack/solid-query'
+import { createMemoryHistory, MemoryRouter } from '@solidjs/router'
 
 const DEMO_LOG_ID = '000000dd--455f14369d'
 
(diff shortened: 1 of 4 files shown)


src/App.browser.test.tsx
```diff
@@ -1,13 +1,36 @@
+import { VoidComponent } from 'solid-js'
```
greatgitsby, line 1: this would be broken out into a separate PR

sshane:
The timeline still flashes for a few frames when clicking routes. There shouldn't be any resource usage there as we already have the data

greatgitsby:
there is still resource usage in this file, events data requires extra requests even if we have the Route; that's why it flashes

this PR was meant to migrate only the `Route` usage

greatgitsby:
@sshane try this version: https://github.com/commaai/connect/pull/539

PR #519, opened 2025-04-08
refactor: delete CircularProgress

greatgitsby:
use the Spinner pattern we use elsewhere, free up the lines!

## mounted usage of the changed components for visual

https://github.com/user-attachments/assets/5ccb7b24-a294-4345-80dc-c0536e6074dd


src/components/material/CircularProgress.tsx
```diff
@@ -1,72 +0,0 @@
...
-      }
-    }
-  }
-
-  return (
-    <span
-      class={clsx(
-        'inline-block',
-        colorClass(),
-        props.progress === undefined ? 'animate-circular-rotate' : 'transition-transform',
```
incognitojam, line 48: We can probably also delete `circular-rotate` and `circular-dash` animations from `tailwind.config.ts`, if there are no other usages

incognitojam (approved):

PR #520, opened 2025-04-08
fix: refresh device list on pair

greatgitsby:
the device list doesn't refresh when you finish pairing your device. now, we refetch the list upon a successful pair


incognitojam:
I was worried that this would be tricky... Nice that it's clean :)

PR #528, opened 2025-04-08
feat(statistics): get route duration from log events

### LINKED ISSUE #247: Stats are way off only for one drive
- Date(3236) and engagement(1818%) is way off
- This route ends up always being on top of my list
- Does not show up on my old connect only on new-connect

Image

incognitojam:
- get route duration from log events (instead of relying on time)
- use new route duration for timeline and engaged % (fixes jitter)
- we still use reported route duration (from route table) to show in route stats sooner, while we fetch the events

closes #247 and fixes the bug with flickering engaged %

### DIFF
--- a/src/api/derived.ts
+++ b/src/api/derived.ts
@@ -184,7 +184,11 @@ const generateTimelineEvents = (routeDurationMs: number, events: DriveEvent[]): 
   return res
 }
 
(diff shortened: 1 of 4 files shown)


sshane:
Will this break for very old routes without events.json? @adeebshihadeh did you just want to use the maxqlog as the duration? That's from the route table and fast

sshane:
Round up to 1m

PR #530, opened 2025-04-08
feat(statistics): use log duration in route stats

incognitojam:
Use route duration from logs for statistics, if available

after #528

### DIFF
--- a/src/components/RouteStatisticsBar.tsx
+++ b/src/components/RouteStatisticsBar.tsx
@@ -1,4 +1,4 @@
-import type { VoidComponent } from 'solid-js'
+import type { Resource, VoidComponent } from 'solid-js'
 
 import type { RouteStatistics } from '~/api/derived'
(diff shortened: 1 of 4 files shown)


incognitojam:
for some reason the duration waits for engaged % to load now! :melting_face:

incognitojam:
reading `data.latest` seems to cause it to wait, too? but checking `data.state === 'ready'` before accessing `data.latest` works OK

I don't understand this, the closest thing mentioned in the docs is this:
<https://docs.solidjs.com/reference/basic-reactivity/create-resource>
image

incognitojam:
When you do `data.latest`, if it's not resolved yet it calls `read()` as though you did `data()`
https://github.com/solidjs/solid/blob/41fa6c14b4bf71eed593303cd63b32d53e3515e9/packages/solid/src/reactive/signal.ts#L739

sshane:
I think it just returns undefined

Issue #463, opened 2025-04-03
Improve font sizes and contrast


sshane:
More contrast, font sizes are too small and hard to read.

PR #542, opened 2025-04-09
fix: rm smallest font size, use consistent font sizes across app

incognitojam:
Define a consistent set of fonts (`text-xs`...`text-2xl`) and remove usages of the smallest font size

Keep around all of the font size definitions as some are used in the material components

#463


sshane:
Can you make the date headings bigger? They're harder to read. Other than that lgtm

PR #553, opened 2025-04-12
fix: speed up drawer open + open route

greatgitsby:
based on feedback from last dev call, speeding up the transitions, tweaking animation to be more like old connect

https://github.com/user-attachments/assets/7db41d0b-76b0-4369-af8b-8ab01acb1028

### DIFF
--- a/src/components/RouteActions.tsx
+++ b/src/components/RouteActions.tsx
@@ -24,7 +24,7 @@ const ToggleButton: VoidComponent<{
       }`}
     >
       <div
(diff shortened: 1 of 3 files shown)


sshane:
225 doesn't exist so it was likely using a shorter default. 300 seems good

https://github.com/tailwindlabs/tailwindcss/blob/d801d8dc5492efec5b830188c2061e31b0379470/packages/tailwindcss/src/compat/default-theme.ts#L1005

PR #566, opened 2025-04-13
fix(UploadQueue): hide 'cancel all' when online queue is empty

### LINKED ISSUE #322: uploading: hide "clear all" button if no online queue items are visible
the UX is poor, X can read as "close," which looks like close the drawer

incognitojam:
Only show the "cancel all" button if the online queue is populated (this should mean the device is online too?)

closes #322

### DIFF
--- a/src/components/UploadQueue.tsx
+++ b/src/components/UploadQueue.tsx
@@ -118,9 +118,11 @@ const UploadQueue: VoidComponent<{ dongleId: string }> = (props) => {
     <div class="flex flex-col gap-4 bg-surface-container-lowest">
       <div class="flex p-4 justify-between items-center border-b-2 border-b-surface-container-low">
         <StatisticBar statistics={[{ label: 'Queued', value: () => items.length }]} />
-        <Button onClick={cancelAll} class="px-2 md:px-3" leading={<Icon name="close" size="20" />} color="primary">
-          <span class="flex items-center gap-1 font-mono">Cancel All</span>
-        </Button>
+        <Show when={onlineQueue.data?.result?.length}>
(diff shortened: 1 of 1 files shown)


incognitojam:
@greatgitsby can you think of any problems with this?

greatgitsby:
overall i think it is fine, it handles:
1. if device is offline, button is not shown
2. if device is online, button is only shown if there are items in the queue

i can think of these two edge cases:
1. if call to online queue fails, it may switch between "device offline" to showing items. button will flash and offline message appears (this could already happen)
2. if there are items currently uploading and nothing backlogged, cancel all button will be there but won't actually do anything

src/components/UploadQueue.tsx
```diff
@@ -118,9 +118,11 @@ const UploadQueue: VoidComponent<{ dongleId: string }> = (props) => {
     <div class="flex flex-col gap-4 bg-surface-container-lowest">
       <div class="flex p-4 justify-between items-center border-b-2 border-b-surface-container-low">
         <StatisticBar statistics={[{ label: 'Queued', value: () => items.length }]} />
-        <Button onClick={cancelAll} class="px-2 md:px-3" leading={<Icon name="close" size="20" />} color="primary">
-          <span class="flex items-center gap-1 font-mono">Cancel All</span>
-        </Button>
+        <Show when={onlineQueue.data?.result?.length}>
```
greatgitsby, line 121:
```suggestion
        <Show when={onlineQueue.data?.result?.filter((item) => !item.current).length}>
```

sshane:
If it adds decent complexity for the edge cases, consider what current connect does and match that for simplification

PR #568, opened 2025-04-13
fix: don't show "no devices" until API response loaded

### LINKED ISSUE #372: Improve first pairing experience
Image

incognitojam:
Don't show "no devices" when loading dashboard - only show when API returned 0 devices

This also removes the initial value `[]` for devices, introduced in [`696b262` (#434)](https://github.com/commaai/connect/pull/434/commits/696b2623326119d42e5c1b1727f6264eb26096a9)
https://github.com/commaai/connect/blob/316512f4beac7fadd9ebb1899ff918681181ea14/src/pages/dashboard/Dashboard.tsx#L110


src/pages/dashboard/Dashboard.tsx
```diff
@@ -92,6 +92,42 @@ const DashboardLayout: Component<{
...
+        leading={
+          <Show when={!modal()} fallback={<DrawerToggleButton />}>
+            <img alt="" src="/images/comma-white.png" class="h-8" />
+          </Show>
+        }
+      >
+        connect
+      </TopAppBar>
+      <div class="flex flex-col gap-4 py-2 items-center mx-auto max-w-sm mt-4 sm:mt-8 md:mt-16">
+        <h2 class="text-xl">Pair your device</h2>
```
greatgitsby, line 110: i think we've preferred `span` over specific header size.

src/pages/dashboard/components/DeviceList.tsx
```diff
@@ -26,7 +26,7 @@ const DeviceList: VoidComponent<DeviceListProps> = (props) => {
   return (
     <List variant="nav" class={props.class}>
       <Suspense fallback={<div class="h-14 skeleton-loader rounded-xl" />}>
-        <For each={props.devices}>
+        <For each={props.devices} fallback={<div class="text-md mx-2 text-on-surface-variant">No devices found</div>}>
```
greatgitsby, line 29: `span`

src/pages/dashboard/Dashboard.tsx
incognitojam: actually, we should probably use more semantic HTML

src/pages/dashboard/Dashboard.tsx
```diff
@@ -106,7 +106,7 @@ const Dashboard: Component<RouteSectionProps> = () => {
     }
   })
 
-  const [devices, { refetch }] = createResource(getDevices, { initialValue: [] })
+  const [devices, { refetch }] = createResource(getDevices, { initialValue: undefined })
```
greatgitsby, line 109: can't we just avoid setting initialValue?

incognitojam: by setting `initialValue` the resource will be immediately resolved and won't trigger Suspense on access

I think without setting this all of the requests will waterfall because we are waiting for devices to resolve before rendering anything else?

Issue #169, opened 2025-02-06
Device settings


incognitojam:
- [x] Update device name
- [x] Manage prime subscription
- [x] Unpair device
- [ ] Device sharing
  - Add or remove other users read only access to the device (current connect doesn't even let you remove users!)

sshane:
I'll consider just the device name changing a requirement for 1.0

incognitojam:
Created a new issue for sharing #580

PR #576, opened 2025-04-14
feat: update device name

incognitojam:
- add material design TextField component
- update device name (alias)

progress towards #169


sshane:
Radius looks disjointed. Can we make them all match? And maybe only show Update button when a change is made, like old connect?

image

Issue #450, opened 2025-04-03
Re-do device query removal


sshane:
Reverted since a PR before it broke some fallback

https://github.com/commaai/connect/pull/435

PR #579, opened 2025-04-14
refactor(Dashboard): shared data

incognitojam:
This removes a couple of duplicate API fetches for profile, devices and the current device. If the current device is in the list already returned (my devices) then that will be used instead, too.

19 requests to api.comma.ai down from 23 (including OPTIONS) on the demo route
https://579.connect-d5y.pages.dev/1d3dc3e03047b0c7/2023-07-27--13-01-19


incognitojam:
Related to #435, #450

PR #591, opened 2025-04-15
feat(Device): add battery voltage indicator

### LINKED ISSUE #433: Add online status and car battery to device stats

incognitojam:
closes #433

| online | loading | offline |
| --- | --- | --- |
| image | image | image |

For an extra 10 lines we can cache the results between changing device, but not sure it's worth it for this specific thing - not sure whether we are trying to replace tanstack or not


sshane:
Why don't we show offline immediately?

incognitojam:
sure we can if the device is fetched already

sshane:
We already show the grey dot

incognitojam:
We wait for the device response before showing the online/offline dot, it's behind a Suspense

This is firing off a request to athena straight away using the dongle ID instead of waiting for that

Can make it wait for the device instead

PR #597, opened 2025-04-15
fix(store): check devices resolved before access

incognitojam:
Avoid using the device list if it's in an errored state


incognitojam:
I tried spam refreshing in Chrome with the dev tools open:
- sometimes both the device list fetch, device fetch and route list fetch all 503
- if only the device list fetch 503s, the demo device still loads and the route list still loads

image

PR #598, opened 2025-04-16
perf(store): wait for device list fetch if pending

incognitojam:
Previously, when requesting a device, we checked if it was already present in the device list to avoid redundant fetches. However, on initial page load, the device list might not have been fetched yet, causing an unnecessary fetch regardless of whether the device is eventually found in the list.

This change ensures we wait for the device list to resolve if it's in a pending or unresolved state to reduce redundant fetches. There is a minor trade-off: a slight delay in fetching devices not in the list on initial load. However, in other cases where the list is already resolved, the fetch is immediate.


incognitojam:
Tested and seems to work, but the Suspense behaviour in DeviceActivity seems different, specifically the online indicator and device name aren't suspended

PR #601, opened 2025-05-06
feat: Device sharing

paytontech:
adds device sharing & un-sharing device as well as a list of users the device is shared with. here's a demo video:

https://github.com/user-attachments/assets/47ce6e55-f9fc-4c7b-8f88-419f5e44c98a

open to feedback, would like to know what to change/fix etc. thanks!

### DIFF
--- a/src/pages/dashboard/activities/SettingsActivity.tsx
+++ b/src/pages/dashboard/activities/SettingsActivity.tsx
@@ -3,7 +3,7 @@ import type { Accessor, VoidComponent, Setter, ParentComponent, Resource, JSXEle
 import { useLocation } from '@solidjs/router'
 import clsx from 'clsx'
 
-import { getDevice, unpairDevice } from '~/api/devices'
+import { getDevice, getDeviceUsers, shareDevice, unpairDevice } from '~/api/devices'
 import {
   cancelSubscription,
   getStripeCheckout,
@@ -22,6 +22,7 @@ import IconButton from '~/components/material/IconButton'
 import TopAppBar from '~/components/material/TopAppBar'
 import { createQuery } from '~/utils/createQuery'
 import { getDeviceName } from '~/utils/device'
+import TextField from '~/components/material/TextField'
 
 const useAction = <T,>(action: () => Promise<T>): [() => void, Resource<T>] => {
   const [source, setSource] = createSignal(false)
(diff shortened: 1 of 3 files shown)


paytontech:
thanks for the feedback! I didn't even notice those APIs were already available, thanks for catching that! I tried using onInput already, but it didn't seem to work. I'll try again, maybe I did something wrong. Agree on the design, it definitely needs some improvement. will work on that, appreciate it

incognitojam:
Those methods are in the other comma-api package, but we aren't using that directly because it's written in JS and doesn't have TypeScript type hints

src/components/material/TextField.tsx
```diff
@@ -100,8 +95,10 @@ const TextField: Component<TextFieldProps> = (props) => {
               getStateStyle().input,
               props.label && labelFloating() && 'pt-6 pb-2',
             )}
-            value={inputValue()}
-            onInput={(e) => setInputValue(e.target.value)}
+            value={props.value}
+            onInput={(e) => {
```
incognitojam, line 99: Is this required? we already passthrough all of the props (see line 91) aside from the ones listed in `splitProps` on line 39

src/pages/dashboard/activities/SettingsActivity.tsx
```diff
@@ -401,15 +402,61 @@ const PrimeManage: VoidComponent<{ dongleId: string }> = (props) => {
...
+                </div>
+              </Show>
+            )}
+          </For>
+          <div class="flex items-center gap-2 justify-between">
+            <TextField
+              label="email"
+              id="email-box"
+              value={shareEmail()}
+              onInput={(val: string) => {
```
incognitojam, line 449: If you remove the custom `oninput` prop from `TextField`, then here you can just do `oninput={(ev) => setShareEmail(ev.target.value)}`

src/pages/dashboard/activities/SettingsActivity.tsx
```diff
@@ -401,15 +402,61 @@ const PrimeManage: VoidComponent<{ dongleId: string }> = (props) => {
...
+                  <div>{user.email}</div>
+                  <Button color="error" onClick={() => unshare(user.email)} loading={unshareLoading()}>
+                    <Icon name="delete" />
+                  </Button>
+                </div>
+              </Show>
+            )}
+          </For>
+          <div class="flex items-center gap-2 justify-between">
+            <TextField
```
incognitojam, line 445: Make the text field fill the entire space instead of leaving a gap between it and the share button

incognitojam (changes requested):
It looks better, but I think it needs a little more work

The device settings didn't look nice in the first place, but you can see in this PR (since reverted, will be added again eventually) where I added the device name field and made the text field take up all of the space, with the update button at the end.

https://github.com/commaai/connect/pull/576
https://590.connect-d5y.pages.dev/c1bd6d7f12f35286/settings
image

src/components/material/TextField.tsx
paytontech: are you referring to getting rid of inputValue()? if so, i believe it is. since there was already an onInput handler here, adding an onInput when using the component wouldn't work, as it would be directed to the existing onInput instead of the new one. I could likely get rid of the custom onInput prop entirely and let it be handled by the built-in onInput, that was an oversight on my part, my bad

paytontech:
got it, thank you!

incognitojam:
Oh I think I see the problem you had with adding your own listener - let me have a look at what I did last time

paytontech:
that's really helpful, I appreciate it! working on doing this now

PR #611, opened 2025-09-07
chore: update to Biome v2

incognitojam:
Adds multi-file analysis and type inference for better linting

Used the migration command to update the config, turned on the recommended rules for the new "solid" domain in the linter, and applied automatic fixes for linting and formatting.

The `.editorconfig` max_line_length isn't respected by Biome (it isn't part of the official editorconfig spec) so I also added that to the `biome.jsonc`, which caused some additional formatting changes.


incognitojam:
Looked over all of the code changes and the automatic fixes look fine

PR #617, opened 2026-03-16
Stabilize geocode tests

Mahdi451:
This PR stabilizes the geocode tests by replacing live Mapbox-dependent assertions with deterministic mocked responses. It also keeps the test job timeout at 2 minutes, since recent runs are already brushing up against the old 1-minute limit. Production geocoding behavior is unchanged.

The tests now cover the early return near `[0, 0]`, request construction, normal address and place-name lookups, failure cases, and place-name fallback behavior.

Testing:
- bun run test src/map/geocode.spec.ts --run
- bun run test --run


Mahdi451:
@incognitojam I’m still fairly new here, so any feedback on this PR would be really appreciated.

Mahdi451:
Closing this in favor of the fork-owned integration PR: https://github.com/Mahdi451/new-connect/pull/1 . The combined branch includes the geocode test fixes plus the fork preview changes that will be validated in that preview.

